PR #22, opened 2020-10-23
Added support for changing the prefix key configuration

hkoturap:
Added support for changing the prefix key configuration

### DIFF
--- a/mtda-cli
+++ b/mtda-cli
@@ -108,10 +108,13 @@ class Application:
 
         client.console_init()
 
+        # Get the value of prefix key
+        prefix_key = client.console_prefix_key()
+
         # Input loop
         while self.exiting is False:
             c = client.console_getkey()
-            if c == '\x01':
+            if c == prefix_key:
                 c = client.console_getkey()
                 self.console_menukey(c)
             else:
--- a/mtda.ini
+++ b/mtda.ini
@@ -24,6 +24,17 @@ console = 5557
 host    = localhost
 
 # ---------------------------------------------------------------------------
+# UI settings
+# ---------------------------------------------------------------------------
+# Set "prefix" to the key which should be followed by command key for
+# controling from interactive console.Default value used for prefix is
+# ctrl-a Note:Supported are values between ctrl-a to ctrl-z with
+# exception of ctrl-m
+# ---------------------------------------------------------------------------
+[ui]
+# prefix = ctrl-a
+
+# ---------------------------------------------------------------------------
 # Console settings
 # ---------------------------------------------------------------------------
 # Set "variant" to specify how to access the console interface of the device.
@@ -77,7 +88,7 @@ variant=aviosys_8800
 # Set "variant" to specify which power control device to use. Use one of:
 #    - qemu
 #    - samsung
-#    - usbf 
+#    - usbf
 # ---------------------------------------------------------------------------
 # Note: this section is ignored when connecting to a remote agent
 # ---------------------------------------------------------------------------
--- a/mtda/main.py
+++ b/mtda/main.py
@@ -24,6 +24,7 @@ _NOPRINT_TRANS_TABLE = {
     i: '.' for i in range(0, sys.maxunicode + 1) if not chr(i).isprintable()
 }
 
+DEFAULT_PREFIX_KEY = 'ctrl-a'
 
 def _make_printable(s):
     return s.translate(_NOPRINT_TRANS_TABLE)
@@ -56,6 +57,7 @@ class MentorTestDeviceAgent:
         self.usb_switches = []
         self.ctrlport = 5556
         self.conport = 5557
+        self.prefix_key = None
         self.is_remote = False
         self.is_server = False
         self.remote = None
@@ -86,6 +88,26 @@ class MentorTestDeviceAgent:
 
         self.mtda.debug(3, "main.command(): %s" % str(result))
         return result
+
+    def console_prefix_key(self):
+        self.mtda.debug(3, "main.console_prefix_key()")
+        return self.prefix_key
+
+    def _prefix_key_code(self, prefix_key):
+        prefix_key = prefix_key.lower()
+        key_dict = {'ctrl-a' : '\x01' , 'ctrl-b' : '\x02' , 'ctrl-c' : '\x03' , 'ctrl-d' : '\x04',
+                    'ctrl-e' : '\x05' , 'ctrl-f' : '\x06' , 'ctrl-g' : '\x07' , 'ctrl-h' : '\x08',
+                    'ctrl-i' : '\x09' , 'ctrl-j' : '\x0A' , 'ctrl-k' : '\x0B' , 'ctrl-l' : '\x0C',
+                    'ctrl-n' : '\x0E' , 'ctrl-o' : '\x0F' , 'ctrl-p' : '\x10' , 'ctrl-q' : '\x11',
+                    'ctrl-r' : '\x12' , 'ctrl-s' : '\x13' , 'ctrl-t' : '\x14' , 'ctrl-u' : '\x15',
+                    'ctrl-v' : '\x16' , 'ctrl-w' : '\x17' , 'ctrl-x' : '\x18' , 'ctrl-y' : '\x19',
+                    'ctrl-z' : '\x1A' }
+
+        if prefix_key in key_dict:
+            key_ascii = key_dict[prefix_key]
+            return key_ascii
+        else:
+            raise ValueError("the prefix key specified %s is not supported" %(prefix_key))
 
     def console_getkey(self):
         self.mtda.debug(3, "main.console_getkey()")
@@ -868,6 +890,8 @@ class MentorTestDeviceAgent:
             self.load_environment(parser)
         if parser.has_section('remote'):
             self.load_remote_config(parser)
+        if parser.has_section('ui'):
+            self.load_ui_config(parser)
         if self.is_remote is False:
             if parser.has_section('power'):
                 self.load_power_config(parser)
@@ -995,6 +1019,10 @@ class MentorTestDeviceAgent:
         else:
             self.remote = None
         self.is_remote = self.remote is not None
+
+    def load_ui_config(self, parser):
+        self.mtda.debug(3, "main.load_ui_config()")
+        self.prefix_key = self._prefix_key_code(parser.get('ui', 'prefix', fallback=DEFAULT_PREFIX_KEY))
 
     def load_usb_config(self, parser):
         self.mtda.debug(3, "main.load_usb_config()")
--- a/mtda/client.py
+++ b/mtda/client.py
@@ -37,6 +37,9 @@ class Client:
     def agent_version(self):
         return self._impl.agent_version()
 
+    def console_prefix_key(self):
+        return self._impl.console_prefix_key()
+
     def command(self, args):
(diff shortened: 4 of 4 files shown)


hkoturap:
@chombourger ,
By default in BBB build we will not have all fields defined in https://github.com/MentorEmbedded/mtda/blob/master/mtda.ini.  Added newly prefix field under remote section. Request your feedback on same.

mtda-cli
```diff
@@ -108,17 +108,37 @@ class Application:
 
         client.console_init()
 
+        key = client.console_prefixkey()
```
chombourger, line 111: `prefix_key = client.console_prefix_key()`

mtda-cli
```diff
@@ -108,17 +108,37 @@ class Application:
...
+
         # Input loop
         while self.exiting is False:
+            prefix = self.console_getprefix(key)
```
chombourger, line 115: does not appear to be required since we have already checked what the prefix key is (just a few lines above)

mtda-cli
```diff
@@ -108,17 +108,37 @@ class Application:
...
             c = client.console_getkey()
-            if c == '\x01':
+            if c == prefix:
```
chombourger, line 117: `if c == prefix_key:`

mtda-cli
```diff
@@ -108,17 +108,37 @@ class Application:
...
                 c = client.console_getkey()
                 self.console_menukey(c)
             else:
-                server.console_send(c, True)
```
chombourger, line 118: we need to send other keys to the console. this line shouldn't be removed

mtda-cli
```diff
@@ -108,17 +108,37 @@ class Application:
...
+                print("\r\nthe specified prefix key of interactive console is not supported\r\n")
 
         print("\r\nThank you for using MTDA!\r\n\r\n")
 
+    def console_getprefix(self, prefixkey):
```
chombourger, line 125: `def console_set_prefix_key(self, prefix_key):`

mtda-cli
```diff
@@ -108,17 +108,37 @@ class Application:
...
+        print("The control prefix key specified is %s" %prefixkey)
```
chombourger, line 126: remove print

chombourger: add something like `prefix_key = prefix_key.lower()`

mtda-cli
```diff
@@ -108,17 +108,37 @@ class Application:
...
+                    'Ctrl-m' : '\x0D' , 'Ctrl-n' : '\x0E' , 'Ctrl-o' : '\x0F' , 'Ctrl-p' : '\x10',
+                    'Ctrl-q' : '\x11' , 'Ctrl-r' : '\x12' , 'Ctrl-s' : '\x13' , 'Ctrl-t' : '\x14',
+                    'Ctrl-u' : '\x15' , 'Ctrl-v' : '\x16' , 'Ctrl-w' : '\x17' , 'Ctrl-x' : '\x18',
+                    'Ctrl-y' : '\x19' , 'Ctrl-z' : '\x1A' }
+
+        if prefixkey in key_dict:
+            key = key_dict[prefixkey]
+            return key
+        else:
+            print("\r\nthe prefix key is not supported\r\n")
```
chombourger, line 139: raise a ValueError instead. message should include the specified `prefix_key`

mtda-cli
```diff
@@ -108,17 +108,37 @@ class Application:
...
                 self.console_menukey(c)
             else:
-                server.console_send(c, True)
+                print("\r\nthe specified prefix key of interactive console is not supported\r\n")
 
         print("\r\nThank you for using MTDA!\r\n\r\n")
 
+    def console_getprefix(self, prefixkey):
+        print("The control prefix key specified is %s" %prefixkey)
+        key_dict = {'Ctrl-a' : '\x01' , 'Ctrl-b' : '\x02' , 'Ctrl-c' : '\x03' , 'Ctrl-d' : '\x04',
```
chombourger, line 127: here we should lower case all valid ctrl keys (e.g. `Ctrl-a` -> 'ctrl-a')
we should also remove keys that conflict with other keys (e.g. `Ctrl-m` is also the return key)

mtda-cli
chombourger: move of the above code should go in a `load_ui_config()` method to be loaded from `load_config` when a `ui` section is found

mtda.ini
```diff
@@ -15,13 +15,16 @@ fuse = no
...
+# Set "prefix" to the key which should be followed by command key for controling from interactive console.
+# Default value used for prefix is Ctrl-a Note:Supported are values between Ctrl-a to Ctrl-z
 # ---------------------------------------------------------------------------
 # Note: the "host" setting is ignored when daemonized
 # ---------------------------------------------------------------------------
 [remote]
 control = 5556
 console = 5557
 host    = localhost
+# prefix = Ctrl-a
```
chombourger, line 27: should really be added to a `ui` section

mtda/main.py
```diff
@@ -56,6 +56,7 @@ def __init__(self):
         self.usb_switches = []
         self.ctrlport = 5556
         self.conport = 5557
+        self.ctrlprefix = 'Ctrl-a'
```
chombourger, line 59: self.prefix_key for consistency

chombourger (changes requested):
code to be restructured per comments

mtda-cli
hkoturap: corrected.

mtda-cli
hkoturap: Added comment correctly.Here we get key defined and below we get ascii value corresponding to key defined.

mtda-cli
hkoturap: updated with correct value.

mtda-cli
hkoturap: Corrected.

mtda-cli
hkoturap: Done.

mtda.ini
hkoturap: Added a new section in ini file for ui section and added prefix key under it.

mtda/main.py
hkoturap: corrected

mtda-cli
hkoturap: @chombourger In case of these keys like 'ctrl-m" we shouldnt support these keys?

mtda-cli
hkoturap: Done

mtda-cli
hkoturap: Done

mtda-cli
chombourger: correct. if you type ctrl-m in your terminal, you will see that it is interpreted as the return key

mtda-cli
```diff
@@ -108,17 +108,39 @@ class Application:
...
-            if c == '\x01':
+            if c == prefix_ascii:
                 c = client.console_getkey()
                 self.console_menukey(c)
             else:
                 server.console_send(c, True)
 
         print("\r\nThank you for using MTDA!\r\n\r\n")
 
+    def console_set_prefix_key(self, prefix_key):
```
chombourger, line 127: we don't really need to expose this method to the outside world.  it is also not setting anything
let's change this to: `def _prefix_key_code(self, prefix_key)`

mtda/main.py
```diff
@@ -996,6 +1003,13 @@ def load_remote_config(self, parser):
             self.remote = None
         self.is_remote = self.remote is not None
 
+    def load_ui_config(self, parser):
+        self.mtda.debug(3, "main.load_ui_config()")
+
+        self.prefix_key = (
+            parser.get('ui', 'prefix', fallback=self.prefix_key))
```
chombourger, line 1010:
```
self.prefix_key = self._prefix_key_code(parser.get('ui', 'prefix', fallback=DEFAULT_PREFIX_KEY))
```

mtda/main.py
```diff
@@ -56,6 +56,7 @@ def __init__(self):
         self.usb_switches = []
         self.ctrlport = 5556
         self.conport = 5557
+        self.prefix_key = 'ctrl-a'
```
chombourger, line 59:
```
self.prefix_key = self._prefix_key_code(DEFAULT_PREFIX_KEY)
```

mtda.ini
```diff
@@ -15,6 +15,8 @@ fuse = no
 # Set "control" to the TCP/IP port number for the control interface
 # Set "console" to the TCP/IP port number for the console I/O interface
 # Set "host" to the remote agent IP address or hostname
+# Set "prefix" to the key which should be followed by command key for controling from interactive console.
```
chombourger, line 18: add line feeds so we do not exceed the dotted line below

mtda.ini
```diff
@@ -23,6 +25,15 @@ control = 5556
 console = 5557
 host    = localhost
 
+# ---------------------------------------------------------------------------
+# UI settings
+# ---------------------------------------------------------------------------
+# Set "prefix" to the key which should be followed by command key for controling from interactive console.
```
chombourger, line 31: add line feeds so we do not exceed the dotted line below

mtda/main.py
```diff
@@ -56,6 +56,8 @@ def __init__(self):
         self.usb_switches = []
         self.ctrlport = 5556
         self.conport = 5557
+        self.default_prefix_key = 'ctrl-a'
```
chombourger, line 59: should declared at the top of this file as a global
```
DEFAULT_PREFIX_KEY = 'ctrl-a'
```

mtda-cli
hkoturap: Moved it to _prefix_key_code under main.py

mtda/main.py
```diff
@@ -996,6 +1020,13 @@ def load_remote_config(self, parser):
             self.remote = None
         self.is_remote = self.remote is not None
 
+    def load_ui_config(self, parser):
+        self.mtda.debug(3, "main.load_ui_config()")
+
+        self.prefix_key = self._prefix_key_code(parser.get('ui', 'prefix', fallback=self.default_prefix_key))
+
+        self.is_remote = self.remote is not None
```
chombourger, line 1028: copy & paste error? shouldn't be needed

mtda-cli
hkoturap: Presently have removed ctrl-m

mtda-cli
hkoturap: Changed it and moved it to main.py

mtda/main.py
hkoturap: Done

mtda/main.py
hkoturap: As we initialize self.prefix_key with fallback value in-case of ui section not defined, presently have not called again _prefix_key_code. Should this be corrected?

chombourger: I believe this would be more correct because this member variable would never hold the string describing the prefix key but rather its ASCII code

mtda.ini
hkoturap: Done

mtda/main.py
hkoturap: Done

mtda/main.py
hkoturap: Corrected

mtda.ini
hkoturap: Done

chombourger (approved):

PR #73, opened 2021-11-24
Initial GitHub workflow

chombourger:


kas-container
```diff
@@ -0,0 +1,454 @@
+#!/bin/sh
```
vj-kumar, line 1: Wondering if we should commit kas-container to this repository?

Can we specify the container to run, like in gitab(`image: ghcr.io/siemens/kas/kas-isar`)?

Or maybe wget it directly in CI.

chombourger: I am unfortunately not up-to-speed with GH actions, this PR was inspired from siemens/meta-iot2050

jan-kiszka: There are basically two recommended patterns here:
- document that the user should fetch a _specific_ version of kas-container and use that (https://source.denx.de/Xenomai/xenomai-images/-/blob/master/README.md#building-target-images)
- embed kas-container in the desired version in your repo (https://github.com/siemens/jailhouse-images)

Originally, I proposed to simply track kas master for the script, but that is risky, specifically when we bump the underlying Debian version for the image.

vj-kumar (approved):

kas-container
chombourger: thank you @jan-kiszka for the guidance!

PR #93, opened 2021-12-02
mtda: bump version to 0.13

chombourger:

### DIFF
--- a/mtda/__version__.py
+++ b/mtda/__version__.py
@@ -11,4 +11,4 @@
 
 __license__ = 'MIT'
 __copyright__ = 'Copyright (C) 2021 Siemens Digital Industries Software'
-__version__ = '0.12'
+__version__ = '0.13'
--- a/debian/changelog
+++ b/debian/changelog
@@ -1,3 +1,22 @@
+mtda (0.13-1) testing; urgency=low
+
+  [Cedric Hombourger]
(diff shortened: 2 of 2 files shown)


chombourger:
See https://groups.google.com/g/mtda-users/c/5dt2fwZ_sIQ
if we feel we should add more changes to master before we release 0.13 then we should discuss there

vj-kumar:
> See https://groups.google.com/g/mtda-users/c/5dt2fwZ_sIQ if we feel we should add more changes to master before we release 0.13 then we should discuss there

I have no changes. 0.13 is to go from my end.

mtda/__version__.py
```diff
@@ -11,4 +11,4 @@
 
 __license__ = 'MIT'
 __copyright__ = 'Copyright (C) 2021 Siemens Digital Industries Software'
-__version__ = '0.12'
+__version__ = '0.13'
```
vj-kumar, line 14: We might need the deb changelog entry as well.

chombourger: had a doubt about the recipe getting debian/changelog from HEAD instead of SRCREV. It is the latter (as expected I guess). adding the `debian/changelog` file I had prepared but not yet staged

vj-kumar (approved):

Issue #67, opened 2021-11-17
mtda-cli help does not (always) return


chombourger:
mtda-cli help does not return. the process appears to be stuck on a lock

chombourger:
we may want to use the python debugger to check threads that are still running. we probably need to stop them

chombourger:
does not seem to be reproducible now with 0.14

chombourger:
reopening, Hemanth was able to reproduce with 0.14 - looking forward to getting reproduction steps

hkoturap:
When running any commands on mtda-assit board, we can observe it does not retun
```
mtda@mtda:~$ sudo mtda-cli help
sudo: unable to resolve host mtda: No address associated with hostname
Device Found
  type: 16c0 05df
  path: /dev/hidraw0
  serial_number: 
  Manufacturer: www.dcttech.com
  Product:      USBRelay2
  Release:      100
  Interface:    0
  Number of Relays =  #2

usage: mtda [options] <command> [<args>]

The most commonly used mtda commands are:
   command   Send a command (string) to the device
   console   Interact with the device console
   getenv    Get named variable from the environment
   monitor   Interact with the device monitor console (if any)
   target    Power control the device
   setenv    Set named variable to specified value in the environment
   storage   Interact with the shared storage device
   usb       Control USB devices attached to the device

Device Found
  type: 16c0 05df
  path: /dev/hidraw0
  serial_number: 
  Manufacturer: www.dcttech.com
  Product:      USBRelay2
  Release:      100
  Interface:    0
  Number of Relays = 2
```
When it doesn't return when we Press Ctrl+C we can observe the following

```
^CException ignored in: <module 'threading' from '/usr/lib/python3.9/threading.py'>
Traceback (most recent call last):
  File "/usr/lib/python3.9/threading.py", line 1428, in _shutdown
    lock.acquire()
KeyboardInterrupt:
```
MTDA version on Assist Board is

```
mtda@mtda:~$ sudo mtda-cli --version
sudo: unable to resolve host mtda: No address associated with hostname
MTDA version: 0.14
```

chombourger:
can you provide your `/etc/mtda/config` file for completeness?

hkoturap:
Following is ``` /etc/mtda/config``` file contents
```
mtda@mtda:~$ cat /etc/mtda/config 
[main]                                                                          
debug=0                                                                        
                                              
[environment]     
variant=ipc427g       

[storage]
variant=usbf
[...]

PR #94, opened 2021-12-02
mtda/main: Stop self._session_timer thread before exiting

bhargavdas:
Currently the self._session_timer thread is not stopped while exiting hence the following error is observe while invoking `ctrl+c` 
```
^CException ignored in: <module 'threading' from '/usr/lib/python3.9/threading.py'>
Traceback (most recent call last):
  File "/usr/lib/python3.9/threading.py", line 1428, in _shutdown
    lock.acquire()
KeyboardInterrupt:
```
This PR stops the thread and exits the application without any errors while using `ctrl+c`

I have also added exit message
```
^CExiting, Thank you for using MTDA ...
```

This PR also fixes [https://github.com/siemens/mtda/issues/67](url) partially.

### DIFF
--- a/mtda/main.py
+++ b/mtda/main.py
@@ -1337,7 +1337,7 @@ class MultiTenantDeviceAccess:
 
         # Stop the timer thread on ctrl+C
         def signal_handler(signum, frame):
-            print("Exiting, Thank you for using MTDA ...")
+            self.mtda.debug(2, "process interrupted, shutting down...")
             self._session_timer.cancel()
 
         signal.signal(signal.SIGINT, signal_handler)
--- a/mtda/main.py
+++ b/mtda/main.py
@@ -1335,6 +1335,12 @@ class MultiTenantDeviceAccess:
         self._session_timer = mtda.utils.RepeatTimer(60, self._session_check)
         self._session_timer.start()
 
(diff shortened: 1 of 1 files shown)


vj-kumar (approved):

mtda/main.py
```diff
@@ -1335,6 +1335,12 @@ def start(self):
         self._session_timer = mtda.utils.RepeatTimer(60, self._session_check)
         self._session_timer.start()
 
+        # Stop the timer thread on ctrl+C
+        def signal_handler(signum, frame):
+            print("Exiting, Thank you for using MTDA ...")
```
chombourger, line 1340: I would not print a message here (or maybe only a debug message using mtda.debug())
user-visible messages should come from mtda-cli

would you agree @vj-kumar ?

other than that, it is a good (and welcome) change

bhargavdas: Ok I will make the message visible through mtda.debug()

@chombourger what will be the ideal debug level 1 or 3 ?

chombourger: we could go with `2` and the message read something like: "process interrupted, shutting down..."

bhargavdas: done

Issue #76, opened 2021-11-29
starting qemu seems to be underdocumented


GuBi34:
Following the documentation on building and starting mtda with QEMU I could not find a link/getting started about how to start the dispatcher utilizing the sample config?

chombourger:
have you checked https://mtda.readthedocs.io/en/latest/build.html#kvm ?

GuBi34:
Hello!

I tried to follow the documentation you mentioned.

* I could build mtda for nano-pi and qemu-amd64 without any issues
* I could start mtda-cli after resolving some python dependencies (which probably are not depicted in the requirements.txt?)
* I failed at step `mtda-cli -d -n`
      * I understood from the docu that it should create the three disk images automatically on start (which did not happen in my case
      * I manually created `usb-sdmux.img` which brought me a bit further

I could not understand from the docu how the mtda qemu is to be started ... so I assumed this will probably be done by the `mtda-cli -d -n` command (just guessing). This one now hangs in

```
File "/home/siemens/local_repos/mtda/mtda/keyboard/qemu.py", line 34, in probe
     result = self.qemu.variant == "qemu"
AttributeError: 'NoneType' object has no attribute 'variant'
```
I looked around in the documentation to figure out how the qemu image is expected to be started ... but could not find something on it.

Hope that helps further to understand my state?


From there I am lagging

chombourger:
prepared several commits to make it easier to install mtda pre-configured for kvm

for the avoidance of doubt, the qemu backend is to emulate a pc and control it using the same interfaces (power, storage, console).

for instance, I can install my mtda-aware (*) OS installer with:
```
# make sure mtda is running as a service
$ sudo systemctl start mtda
# check if qemu was properly started
$ sudo journalctl -u mtda|tail -n 5
Nov 30 15:48:22 build systemd[1]: Started Multi-Tenant Device Access.
Nov 30 15:48:22 build mtda-cli[209837]: QEMU emulator version 5.2.0 (Debian 1:5.2+dfsg-11)
Nov 30 15:48:22 build mtda-cli[209837]: Copyright (c) 2003-2020 Fabrice Bellard and the QEMU Project developers
Nov 30 15:48:22 build mtda-cli[209835]: # debug2: power.qemu.start(): swtpm process started [209849]
Nov 30 15:48:22 build mtda-cli[209835]: # debug2: power.qemu.start(): qemu process started [209857]
# write my installer image to the shared storage device
$ mtda-cli target off
[...]

PR #96, opened 2021-12-05
Qemu packages

chombourger:


chombourger:
Should also help with #76

bhargavdas:
@chombourger I reviewed the changes/additions everything looks fine, my only comment is as you mentioned these prebuilt packages are meant for debian 11 shouldn't it be mentioned somewhere in the docs. Maybe in prerequisites.

chombourger:
> @chombourger I reviewed the changes/additions everything looks fine, my only comment is as you mentioned these prebuilt packages are meant for debian 11 shouldn't it be mentioned somewhere in the docs. Maybe in prerequisites.

thanks. will think about it. maybe we need multiple repositories in fury (e.g. mtda-bullseye and mtda-bionic)

PR #112, opened 2021-12-20
Prepare support for newer nanopi variants

chombourger:


chombourger:
Gunther indicated that our image actually boots but none of the network interfaces were up.

vj-kumar (approved):

vj-kumar:
Now that this boots. We can work with this as base and debug the network issue.

PR #129, opened 2021-12-30
Add support for wifi and bluetooth

vj-kumar:
Testing wifi:
`
sudo nmcli dev wifi connect <ssid name> password <password>`

Testing bluetooth:
```
bluetoothctl
# scan on
# pair <mac address>
```

### DIFF
--- a/meta-isar/conf/machine/nanopi-r1.conf
+++ b/meta-isar/conf/machine/nanopi-r1.conf
@@ -14,3 +14,5 @@ WKS_FILE ?= "nanopi-r1.wks.in"
 
 IMAGE_INSTALL += "u-boot-script"
 IMAGER_INSTALL += "u-boot-nanopi-r1"
+
+IMAGE_INSTALL += " firmware-nanopi-r1"
--- a/meta-isar/recipes-bsp/firmware-nanopi-r1/firmware-nanopi-r1.bb
+++ b/meta-isar/recipes-bsp/firmware-nanopi-r1/firmware-nanopi-r1.bb
@@ -1,0 +1,16 @@
+# This Isar layer is part of MTDA
+# Copyright (C) 2021 Siemens Digital Industries Software
+# SPDX-License-Identifier: MIT
+
+inherit dpkg-raw
+
+DESCRIPTION = "Wi-Fi firmware suport for NanoPi R1"
+MAINTAINER = "mtda-users <mtda-users@googlegroups.com>"
+DEBIAN_DEPENDS = "firmware-brcm80211"
+
+do_install() {
+    targetdir=${D}/lib/firmware/brcm
+    targetlnk=${targetdir}/brcmfmac43430-sdio.friendlyarm,nanopi-r1.txt
+    install -m 0755 -d ${targetdir}
+    ln -s brcmfmac43430-sdio.AP6212.txt ${targetlnk}
+}
--- a/meta-isar/recipes-core/images/mtda-image.inc
+++ b/meta-isar/recipes-core/images/mtda-image.inc
@@ -38,6 +38,7 @@ IMAGE_PREINSTALL += "                    \
     vim                                  \
     wireless-regdb                       \
     wireless-tools                       \
+    bluetooth                            \
 "
 
 # HomeKit support
--- a/meta-isar/recipes-core/images/mtda-image.inc
+++ b/meta-isar/recipes-core/images/mtda-image.inc
@@ -36,6 +36,8 @@ IMAGE_PREINSTALL += "                    \
     ssh                                  \
     sudo                                 \
     vim                                  \
+    wireless-regdb                       \
+    wireless-tools                       \
 "
 
 # HomeKit support


chombourger:
SPDX headers are missing
Find out which with:
pip3 install --user reuse
reuse lint

meta-isar/recipes-bsp/firmware-nanopi-r1/files/postinst
```diff
@@ -0,0 +1,11 @@
+#!/bin/sh
+
+set -e
+
+NANOPI_NVRAM_CONFIG="/lib/firmware/brcm/brcmfmac43430-sdio.friendlyarm,nanopi-r1.txt"
+AP6212_NVRAM_CONFIG="/lib/firmware/brcm/brcmfmac43430-sdio.AP6212.txt"
+if [ ! -f "${NANOPI_NVRAM_CONFIG}" ] && [ -f "${AP6212_NVRAM_CONFIG}" ]; then
+    ln -sf "${AP6212_NVRAM_CONFIG}" "${NANOPI_NVRAM_CONFIG}"
+    mkdir -p /usr/share/firmware-nanopi-r1/
+    touch /usr/share/firmware-nanopi-r1/symlink
```
chombourger, line 10: Not sure why we need this dummy symlink file?

meta-isar/recipes-bsp/firmware-nanopi-r1/files/prerm
```diff
@@ -0,0 +1,10 @@
+#!/bin/sh
+
+set -e
+
+NANOPI_NVRAM_CONFIG="/lib/firmware/brcm/brcmfmac43430-sdio.friendlyarm,nanopi-r1.txt"
+
+if [ -f /usr/share/firmware-nanopi-r1/symlink ]; then
```
chombourger, line 7: Could be replaced with

if [ -h <mysdio.txt> ]; then
    rm -f <mysdio.txt>
fi

meta-isar/conf/machine/nanopi-r1.conf
```diff
@@ -14,3 +14,6 @@ WKS_FILE ?= "nanopi-r1.wks.in"
 
 IMAGE_INSTALL += "u-boot-script"
 IMAGER_INSTALL += "u-boot-nanopi-r1"
+
+IMAGE_PREINSTALL += " wireless-regdb wireless-tools bluetooth"
```
chombourger, line 18: I would be tempted to add all generic wireless packages to mtda-image.inc

meta-isar/recipes-bsp/firmware-nanopi-r1/files/postinst
vj-kumar: To know if this package has installed the symlink or whether its from upstream.

meta-isar/conf/machine/nanopi-r1.conf
vj-kumar: Sure.

meta-isar/recipes-bsp/firmware-nanopi-r1/files/postinst
vj-kumar: As of date there is no such symlink upstream though.

vj-kumar: I think there is a better way to do this. Let me try that and update the PR

vj-kumar: We can use dpkg-query to make sure that symlink doesnot belong to any other package.

meta-isar/recipes-bsp/firmware-nanopi-r1/firmware-nanopi-r1.bb
```diff
@@ -0,0 +1,13 @@
...
+
+DESCRIPTION = "Wifi firmware symlink for NanoPi R1"
+MAINTAINER = "mtda-users <mtda-users@googlegroups.com>"
+DEBIAN_DEPENDS = "firmware-brcm80211"
+
+SRC_URI = " file://postinst \
+        file://prerm \
+        file://rules"
+
+inherit dpkg-raw
```
chombourger, line 13: A possibly easier way is to ship the symbolic link and let apt report an error in the unlikely event where two packages are shipping `brcmfmac43430-sdio.friendlyarm,nanopi-r1.txt`.

Here's how I reworked your recipe:
```
# Copyright (C) 2021 Siemens Digital Industries Software

inherit dpkg-raw

DESCRIPTION = "Wi-Fi firmware suport for NanoPi R1"
MAINTAINER = "mtda-users <mtda-users@googlegroups.com>"
DEBIAN_DEPENDS = "firmware-brcm80211"

do_install() {
    targetdir=${D}/lib/firmware/brcm
    targetlnk=${targetdir}/brcmfmac43430-sdio.friendlyarm,nanopi-r1.txt
    install -m 0755 -d ${targetdir}
    ln -s brcmfmac43430-sdio.AP6212.txt ${targetlnk}
}
```
We may then remove `postinst`, `prerm` and `rules`

chombourger:
```
root@mtda:/lib/firmware/brcm# ls -l
total 11668
...
-rw-r--r-- 1 root root     875 Jul 25 21:19 brcmfmac43430-sdio.AP6212.txt
lrwxrwxrwx 1 root root      31 Jul 25 21:19 brcmfmac43430-sdio.bin -> ../cypress/cyfmac43430-sdio.bin
lrwxrwxrwx 1 root root      36 Jul 25 21:19 brcmfmac43430-sdio.clm_blob -> ../cypress/cyfmac43430-sdio.clm_blob
lrwxrwxrwx 1 root root      29 Jan  2 08:20 brcmfmac43430-sdio.friendlyarm,nanopi-r1.txt -> brcmfmac43430-sdio.AP6212.txt
...
```

chombourger: and
```
root@mtda:/# dpkg -S /lib/firmware/brcm/brcmfmac43430-sdio.friendlyarm,nanopi-r1.txt 
firmware-nanopi-r1: /lib/firmware/brcm/brcmfmac43430-sdio.friendlyarm,nanopi-r1.txt
```

chombourger:
We should amend the sumary line of both commits: `Add` -> `add`
and `WiFi` -> `Wi-Fi`

```
f125ebd nanopi-r1: add support for Bluetooth
132283d nanopi-r1: add support for Wi-Fi
```

meta-isar/recipes-bsp/firmware-nanopi-r1/firmware-nanopi-r1.bb
vj-kumar: Sure. This also works.

vj-kumar:
> We should amend the sumary line of both commits: `Add` -> `add` and `WiFi` -> `Wi-Fi`
> 
> ```
> f125ebd nanopi-r1: add support for Bluetooth
> 132283d nanopi-r1: add support for Wi-Fi
> ```

Done.

chombourger (approved):

Issue #132, opened 2022-01-04
ModuleNotFoundError: No module named 'zeroconf'


ricbha:
After mtda installation on target, zeroconf module is required to run mtda-cli
Log:
```
# mtda-cli help
Traceback (most recent call last):
  File "/root/.local/bin/mtda-cli", line 26, in <module>
    import zeroconf
ModuleNotFoundError: No module named 'zeroconf'

```

PR #133, opened 2022-01-04
setup.py: needs zeroconf to be installed to run mtda-cli on host

ricbha:
Closes: #132
 

chombourger:
thanks for catching this and for the fix!

Issue #148, opened 2022-01-10
storage: mtda-usb-functions service failure


ricbha:
mtda-usb-functions service fails mostly on reboot with following error:

```
Jan 04 18:29:27 mtda systemd[1]: Started USB functions for MTDA.
Jan 04 18:29:28 mtda usb-functions[312]: FileNotFoundError: [Errno 2] No such f>
Jan 04 18:29:28 mtda usb-functions[312]: During handling of the above exception>
Jan 04 18:29:28 mtda usb-functions[312]: Traceback (most recent call last):
Jan 04 18:29:28 mtda usb-functions[312]:   File "/usr/libexec/mtda/usb-function>
Jan 04 18:29:28 mtda usb-functions[312]:     install()
Jan 04 18:29:28 mtda usb-functions[312]:   File "/usr/libexec/mtda/usb-function>
Jan 04 18:29:28 mtda usb-functions[312]:     fileputcontent(path + "/functions/>
Jan 04 18:29:28 mtda usb-functions[312]:   File "/usr/libexec/mtda/usb-function>
Jan 04 18:29:28 mtda usb-functions[312]:     fp.write(bytearray(content))
Jan 04 18:29:28 mtda usb-functions[312]: FileNotFoundError: [Errno 2] No such f>
Jan 04 18:29:28 mtda systemd[1]: mtda-usb-functions.service: Main process exite>
Jan 04 18:29:28 mtda systemd[1]: mtda-usb-functions.service: Failed with result>
```

Reason for failure is because this service is looking for /dev/sda even before it is loaded
```
[   15.056464] Mass Storage Function, version: 2009/09/11                       
[   15.061874] LUN: removable file: (no medium)                                 
[   15.064769] mass_storage.usb0/lun.0: unable to open backing file: /dev/sda   
[   15.084285] mass_storage.usb0/lun.0: unable to open backing file: /dev/sda         
[   28.634113] scsi 0:0:0:0: Attached scsi generic sg0 type 0                   
[   28.701273] sd 0:0:0:0: [sda] 60678144 512-byte logical blocks: (31.1 GB/28.)
[   28.710382] sd 0:0:0:0: [sda] Write Protect is off                           
[   28.715332] sd 0:0:0:0: [sda] Mode Sense: 45 00 00 00                        
[   28.721297] sd 0:0:0:0: [sda] Write cache: disabled, read cache: enabled, doA
[   29.002663]  sda: sda1 sda2                                                  
[...]

PR #150, opened 2022-01-11
storage: start mtda-usb-functions after storage is loaded

ricbha:
Closes: #148 


ricbha:
@chombourger Should changes be added to https://github.com/siemens/mtda/blob/master/docs/fixtures.rst?

meta-isar/recipes-conf/mass-storage/usb-ms-use-sda/10-after-sda.conf
```diff
@@ -1,4 +1,5 @@
 # Copyright (C) 2021 Siemens Digital Industries Software
 # SPDX-License-Identifier: MIT
 [Unit]
+Wants=dev-sda.device
```
chombourger, line 4: for completeness we should also update the systemd unit for mmcblk0 with the same change

chombourger:
we should also rename `usb-ms-use-sda_0.4.bb` to `usb-ms-use-sda_0.5.bb` and `usb-ms-use-mmcblk0_0.4.bb` to `usb-ms-use-mmcblk0_0.5.bb`

this is needed for the packages on Gemfury to be updated

chombourger:
> @chombourger Should changes be added to https://github.com/siemens/mtda/blob/master/docs/fixtures.rst?

good catch. we should

chombourger (approved):

chombourger:
Looks good to me, thank you so much for your work!

Issue #147, opened 2022-01-10
storage: writer: BZ2 compression broken


vj-kumar:
Flashing BZ2 files is broken.

PR #152, opened 2022-01-12
storage/writer: handle end of stream for bzip2 streams

chombourger:
Closes: #147

### DIFF
--- a/mtda/storage/writer.py
+++ b/mtda/storage/writer.py
@@ -3,7 +3,7 @@
 # ---------------------------------------------------------------------------
 #
 # This software is a part of MTDA.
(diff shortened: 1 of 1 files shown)


chombourger:
@vj-kumar this fixes #147 for me

was the issue you found that `storage write` would never return on the client side and that an `EOFError` was seen on the server side?

vj-kumar:
> @vj-kumar this fixes #147 for me
> 
> was the issue you found that `storage write` would never return on the client side and that an `EOFError` was seen on the server side?

Yes. Exactly.

vj-kumar (approved):

Issue #121, opened 2021-12-23
support xz compressed images


chombourger:
use https://github.com/Rogdham/python-xz to support .xz images

vj-kumar:
Not sure if someone is looking into to it. Planing to take it.

chombourger:
> Not sure if someone is looking into to it. If not planing to take it.

Thanks!

vj-kumar:
@chombourger Any specific reason for using python-xz? Looks like lzma can do it to. I do see some benefits mentioned in python-xz, but currently not sure whether it will have an impact.

jan-kiszka:
If you are talking about isar image compression: Adriaan has started porting over the OE image compression features. That will obsolete the currently needed custom image classes.

vj-kumar:
@jan-kiszka This is about mtda's storage writer supporting xz compressed images. https://github.com/siemens/mtda/blob/a13e98c44271218adfb850f23119a9788234265b/mtda/storage/writer.py#L49-L60

vj-kumar:
Bz2 write is broken. Tested in Qemu, Richa also validated in an actual board. XZ similar implementation to BZ2 also doesnot work. Checking.

chombourger:
Thanks for the info. We will get a framework to test our APIs and features With https://github.com/siemens/mtda/pull/146 and hopefully avoid regressions in the future

vj-kumar:
Related to https://github.com/siemens/mtda/issues/147

Planning to address #147 before getting here.

PR #153, opened 2022-01-12
mtda: add support for xz compressed images

vj-kumar:
Based on bz2 logic.

Closes: #121

Signed-off-by: Vijai Kumar K <Vijaikumar_Kanagarajan@mentor.com>

### DIFF
--- a/mtda/client.py
+++ b/mtda/client.py
@@ -245,6 +245,8 @@ class Client:
                 compression = CONSTS.IMAGE.GZ.value
             elif path.endswith(".zst"):
                 compression = CONSTS.IMAGE.ZST.value
+            elif path.endswith(".xz"):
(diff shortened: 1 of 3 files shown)


vj-kumar:
Tested in qemu.

chombourger:
looks good. we may need to add lzma to setup.py for this new dependency

vj-kumar:
> looks good. we may need to add lzma to setup.py for this new dependency

It is part of the python standard library. https://docs.python.org/3/library/lzma.html

chombourger:
> > looks good. we may need to add lzma to setup.py for this new dependency
> 
> It is part of the python standard library. https://docs.python.org/3/library/lzma.html

ah did not know. thanks for clarifying

PR #160, opened 2022-01-14
meta-isar/conf: introduce mtda-bullseye distro

chombourger:
Having per-machine DISTRO_APT_SOURCES or DISTRO_APT_PREFERENCES
was resulting in differences in the isar-bootstrap-target builds
and would cause problems in multi-config builds. Moreover, the
"abstraction" layer provided by the MTDA_EXTRA_APT variables
introduced in c610fef was pointless. A mtda-bullseye distro is
instead introduced (it extends debian-bullseye) and includes
backports we require for NanoPi variants such as the R1.


kas/common/base.yml
```diff
@@ -0,0 +1,37 @@
...
+
+target: isar-image-base
+
+repos:
+  mtda:
+    layers:
+      meta-isar:
+  isar:
+    url: https://github.com/ilbers/isar.git
+    refspec: master
```
jan-kiszka: Not a good idea. Set this in a nightly CI build or so, but it's obviously nothing for common use. No layer used in production does that.

kas/common/base.yml
```diff
@@ -0,0 +1,37 @@
...
+    layers:
+      meta:
+      meta-isar:
```
jan-kiszka, line 26: Are you sure that you use meta-isar? It's generally for demo/testing/reference purposes.

chombourger: we do - it ships the nanopi-neo machine

jan-kiszka: Ah, ok. As long as you can use those two files, this makes sense.

kas/common/base.yml
chombourger: I removed the use of `master` in the following commit `kas/common: use older Isar to work-around build failures`
I can squash these two if you prefer (I felt I should refactor first and then use a fixed refspec but I am happy to change it with the refactoring itself)

jan-kiszka: Given that `master` was used before, squashing is not actually needed. I would just rephrase the last commit towards "use a well-defined Isar version".

chombourger: done

PR #167, opened 2022-01-18
prepare version 0.15

chombourger:


debian/changelog
```diff
@@ -1,3 +1,43 @@
...
+  * upload .deb packages to Gemfury only for release tags
+  * make on/off functions no-ops when target is already on/off
+  * make on function deliver power before resuming the console
+  * start session timer only when running as server
+  * do not send empty session events
+  * fix "storage write" for raw files
+  * fix "storage write" for bz2 files
+  * fix "power event not being cleared on power off" in several drivers
+
+  [Vijay Kumar K]
```
vj-kumar, line 22: That should be an i. Not a common spelling for that name. :) But i it is. Vijai Kumar K.

vj-kumar: LGTM otherwise. Can be merged. I suppose we are manually creating the changelog? I use  gbp dch.

chombourger: correct manual. should probably check if we should use `gbp dch` moving forward (will then need to make sure all commit logs are "pretty")

chombourger: gave `gbp dch` a try and updated this PR

vj-kumar (approved):

Issue #151, opened 2022-01-11
keyboard/hid: broken pipe


chombourger:
Gunther experienced the following issue with one his IPCs
![image](https://user-images.githubusercontent.com/90115/149009357-7f1db96e-ce7b-474d-bff6-6c858fe02ad9.png)

chombourger:
Here's the power script from his setup:
```
[scripts]
power on:
	variant = 'ipc227e'
	if 'variant' in env:
	... variant = env['variant']
	... mtda.debug(1, "!!!Variant is set")
	if 'boot-from-usb' in env and env['boot-from-usb'] == '1':
	... import time
	... time.sleep(5)
	... mtda.keyboard.probe()
	... mtda.debug(1, "!!!Keyboard is probed")
	... mtda.keyboard.esc(5)
	... mtda.debug(1, "!!!Sent ESC to DUT")
	... time.sleep(1)
	... if variant == 'ipc227e':
	... ... mtda.keyboard.right()
	... ... mtda.debug(1, "!!!Variant is set as ipc227e")
	... elif variant == 'ipc127e':
	... ... mtda.keyboard.down()
	... ... mtda.debug(1, "!!!Variant is set as ipc127e")
	... mtda.keyboard.enter()
	... time.sleep(0.5)
	... mtda.keyboard.down()
	... mtda.keyboard.enter()
	... mtda.keyboard.idle()
```
he then said that using sleep(5) or sleep(6) does not work but sleep(5.5) does

chombourger:
I had experienced a similar issue with OCII platforms because the BIOS would not support the USB composite device we provide and would not respond to the USB reports we send

hkoturap:
Presently with couple of QA Setups we are experiencing the issue of `BrokenPipeError: [Errno 108] Cannot send after transport endpoint shutdown` after switching to 0.14 MTDA  version.
```
# debug3: keyboard.hid.write_report()
Traceback (most recent call last):
  File "/usr/lib/python3/dist-packages/mtda/keyboard/hid.py", line 87, in press
    written = self.write_report(
  File "/usr/lib/python3/dist-packages/mtda/keyboard/hid.py", line 56, in write_report
    return self.fd.write(report.encode())
BrokenPipeError: [Errno 108] Cannot send after transport endpoint shutdown
```

Below is the complete log.
```
mtda@mtda:~$ sudo mtda-cli target on
# debug3: main.load_environment()
# debug4: main.load_environment(): variant => ipc427e
[...]

PR #173, opened 2022-01-20
keyboard/hid: make write_report() non-blocking

chombourger:
Give-up after a second if we cannot write the HID report.

Fixes: #151


chombourger:
Tested on the NanoPI-R1 (running a 5.14 kernel)

bikrambhola (approved):

PR #175, opened 2022-01-21
Qemu configuration for multiple cores

GuBi34:
Provide QEMU configuration for the number of cores

Signed-off-by: Gunther Birk <gunther.birk@siemens.com>

### DIFF
--- a/configs/qemu.ini
+++ b/configs/qemu.ini
@@ -35,6 +35,7 @@ variant=qemu
 [power]
 variant=qemu
 cpu=host
+smp=0
 machine=pc
 memory=2048
 pflash_ro=/usr/share/OVMF/OVMF_CODE.fd
--- a/docs/config.rst
+++ b/docs/config.rst
@@ -276,6 +276,11 @@ The following settings are supported:
 * ``cpu``: string [optional]
     The CPU to be emulated by QEMU/KVM.
 
+* ``smp``: integer [optional]
+    ``smp=<NUMBER>`` specifies the number of cores the guest is permitted to
+    use. The number can be higher than the available cores on the host system.
+    Use ``smp=0`` to use all currently available cores.
+
 * ``executable``: string [optional]
     The QEMU/KVM executable to use as system emulator. This setting defaults
     to ``kvm``
--- a/mtda/power/qemu.py
+++ b/mtda/power/qemu.py
@@ -19,6 +19,7 @@ import signal
 import tempfile
 import threading
 import time
+import multiprocessing
 
 # Local imports
 from mtda.power.controller import PowerController
@@ -31,6 +32,7 @@ class QemuController(PowerController):
         self.ev = threading.Event()
         self.bios = None
         self.cpu = None
+        self.smp = None
         self.drives = []
         self.executable = "kvm"
         self.hostname = "mtda-kvm"
@@ -52,6 +54,8 @@ class QemuController(PowerController):
             self.bios = conf['bios']
         if 'cpu' in conf:
             self.cpu = conf['cpu']
+        if 'smp' in conf:
+            self.smp = int(conf['smp'])
         if 'executable' in conf:
             self.executable = conf['executable']
         if 'hostname' in conf:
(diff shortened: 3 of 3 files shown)


docs/config.rst
```diff
@@ -276,6 +276,9 @@ The following settings are supported:
 * ``cpu``: string [optional]
     The CPU to be emulated by QEMU/KVM.
 
+* ``smp``: string [optional]
+    ``smp=<NUMBER>`` specifies the number of cores the guest is permitted to use. The number can be higher than the available cores on the host system. Use ``smp=$(nproc)`` to use all currently available cores.
```
vj-kumar, line 280: Can you wrap this line at 80 like the rest of the doc?

configs/qemu.ini
```diff
@@ -35,6 +35,7 @@ variant=qemu
 [power]
 variant=qemu
 cpu=host
+smp=$(nproc)
```
chombourger, line 38: This works because we construct the command line verbatim. I would not recommend we advertise the ability to pass shell constructs. Can we say "0" means automatic and have power/qemu run nproc (or even better use some python APIs) to determine how many cores/cpus we have.

chombourger:
```
import multiprocessing
cpus = multiprocessing.cpu_count()
```
on my laptop: this returns `4` (same as `nproc`)

GuBi34: What would be the reasonable default value?!

* no smp configured -> go with 1 Core (no command line entry for qemu)
* `smp=0` -> start qemu with `-smp ...` set to cpu_count()
* `smp=<number>` -> start qemu with `-smp <number>`

chombourger: that sounds good to me!

mtda/power/qemu.py
```diff
@@ -31,7 +32,7 @@ def __init__(self, mtda):
         self.ev = threading.Event()
         self.bios = None
         self.cpu = None
-        self.smp = None
+        self.smp = None
```
chombourger, line 35: we may have a trailing space here (not easy to tell from the GH interface)

docs/config.rst
```diff
@@ -277,7 +277,9 @@ The following settings are supported:
     The CPU to be emulated by QEMU/KVM.
 
 * ``smp``: string [optional]
```
chombourger, line 279: should be integer

chombourger:
make sure you squash the 2 commits together
(git rebase -i origin/master and set the action to `squash` for the 2nd commit)

configs/qemu.ini
chombourger: set to smp=0 here

chombourger (approved):

chombourger:
LGTM, thank you so much
Triggered a CI check

chombourger:
If you don't mind changing the subject line of your commit from;
`power/qemu: Qemu configuration for multiple cores`
to `power/qemu: support configuration for multiple cores`

chombourger:
CI has found the following issues:
```
mtda/power/qemu.py:35:24: W291 trailing whitespace
mtda/power/qemu.py:149:16: F632 use ==/!= to compare constant literals (str, bytes, int, float, tuple)
```
they need to be fixed before we can merge (you can run `flake8` locally to make sure your code is OK)

mtda/power/qemu.py
```diff
@@ -141,6 +145,11 @@ def start(self):
             options += " -bios %s" % self.bios
         if self.cpu is not None:
             options += " -cpu %s" % self.cpu
+        if self.smp is not None:
+            if self.smp is 0:
```
chombourger, line 149: == 0 since this is now an integer

mtda/power/qemu.py
chombourger: looks like the trailing space is still there (according to flake8)

configs/qemu.ini
GuBi34: done

chombourger (approved):

PR #181, opened 2022-01-30
add support for the beaglebone-black

chombourger:
I haven't tried booting the image just yet

### DIFF
--- a/kas/debian/mtda-beaglebone-black.yml
+++ b/kas/debian/mtda-beaglebone-black.yml
@@ -1,0 +1,22 @@
+# ---------------------------------------------------------------------------
+# Debian image to run MTDA on the BeagleBone Black
+# ---------------------------------------------------------------------------
+#
+# This software is a part of MTDA.
+# Copyright (C) 2022 Siemens Digital Industries Software
+#
+# ---------------------------------------------------------------------------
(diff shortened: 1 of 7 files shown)


chombourger:
boot and smoke tested. this PR may now be reviewed

meta-isar/recipes-bsp/u-boot/u-boot-beaglebone-black_2021.10.bb
```diff
@@ -0,0 +1,19 @@
+# ---------------------------------------------------------------------------
```
vj-kumar, line 1: Can't we use the MLO & u-boot.img in /usr/lib/u-boot/am335x_boneblack/ via u-boot-omap package and avoid this recipe?

https://packages.debian.org/bullseye/armhf/u-boot-omap/filelist

chombourger: yes, this version appears to be compatible. did the change locally and the system was booted ok. doing a clean build to double-check

chombourger: build & boot tested clean build: all ok

vj-kumar (approved):

PR #183, opened 2022-02-01
main: start with board off and storage on target

chombourger:
Put test fixture like power controller or storage muxes into a known
state when MTDA is started.

### DIFF
--- a/mtda/main.py
+++ b/mtda/main.py
@@ -1381,6 +1381,10 @@ class MultiTenantDeviceAccess:
                                                          self._session_check)
             self._session_timer.start()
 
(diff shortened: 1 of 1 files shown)


chombourger:
Worked today with QA setup for the IPC427e.  Power is controlled with pduclient. That driver has no way to query LAVA to determine the current state of the power switch. The driver initially reports "???". By making an explicit call to `target_off()` from `start()`, "OFF" will be returned instead and `storage_to_host()` or `storage_to_target()` will not complain about the storage being locked.

vj-kumar (approved):

Issue #184, opened 2022-02-02
web-based access to the console


chombourger:
run a lightweight web server on the assist board and consider using https://xtermjs.org to provide a terminal running in the browser
video stream may also be rendered in the browser when using `mjpg_streamer`

PR #189, opened 2022-02-04
www: allow remote access to the console via web browsers

chombourger:
Make ConsoleLogger send output to a Flask-SocketIO socket that may
be consumed by an xterm.js-based terminal in the browser.

Closes: #184


chombourger:
Have you had a chance to check this PR?

PR #192, opened 2022-02-04
meta-isar/recipes-python: build HEAD of mtda

chombourger:
Make it easier to build MTDA images with the latest changes by
using SRCREV="${AUTOREV}" in the bitbake recipe. Releases will
continue to use a fixed git commit hash. The Debian changelog
uses a "-0" revision to flag development builds.

### DIFF
--- a/debian/changelog
+++ b/debian/changelog
@@ -1,3 +1,9 @@
+mtda (0.16-0) testing; urgency=medium
+
+  * Development version
+
+ -- Cedric Hombourger <cedric.hombourger@siemens.com>  Fri, 04 Feb 2022 08:00:00 +0100
+
 mtda (0.15-1) testing; urgency=medium
 
   [ Cedric Hombourger ]
--- a/meta-isar/recipes-python/mtda/mtda_git.bb
+++ b/meta-isar/recipes-python/mtda/mtda_git.bb
@@ -8,7 +8,7 @@
 inherit dpkg
 
 SRC_URI += "git://git@github.com/siemens/mtda.git;protocol=https;branch=master"
-SRCREV   = "4925f8f1bc1214f78777a7648eac0ac183f11232"
+SRCREV   = "${AUTOREV}"
 S        = "${WORKDIR}/git"
 
 DEPENDS += "zerorpc-python zstandard"
--- a/mtda/__version__.py
+++ b/mtda/__version__.py
@@ -11,4 +11,4 @@
 
 __license__ = 'MIT'
 __copyright__ = 'Copyright (C) 2022 Siemens Digital Industries Software'
-__version__ = '0.15'
+__version__ = '0.16'


vj-kumar:
Can we conditionally build stable? Maybe defaulting to autorev but an option to build stable?

chombourger:
> Can we conditionally build stable? Maybe defaulting to autorev but an option to build stable?

I guess we could but was worried that the end-result would be a mix of stable and devel (since HEAD may contain changes to the image that aren't compatible with "stable"). I would recommend people wanting to build stable to checkout a specific tag (e.g. v0.15) so they get everything aligned. Thoughts?

vj-kumar:
> > Can we conditionally build stable? Maybe defaulting to autorev but an option to build stable?
> 
> I guess we could but was worried that the end-result would be a mix of stable and devel (since HEAD may contain changes to the image that aren't compatible with "stable"). I would recommend people wanting to build stable to checkout a specific tag (e.g. v0.15) so they get everything aligned. Thoughts?

Then again, in the event of release you would modify the changelog, add a fixed commit, tag it and then reintroduce a development version like now(Autorev + dev changelog)?

chombourger:
> > > Can we conditionally build stable? Maybe defaulting to autorev but an option to build stable?
> > 
> > 
> > I guess we could but was worried that the end-result would be a mix of stable and devel (since HEAD may contain changes to the image that aren't compatible with "stable"). I would recommend people wanting to build stable to checkout a specific tag (e.g. v0.15) so they get everything aligned. Thoughts?
> 
> Then again, in the event of release you would modify the changelog, add a fixed commit, tag it and then reintroduce a development version like now(Autorev + dev changelog)?

Yes that's right. We would use a fixed SRCREV when preparing our release (-rcX tags) and switch back to AUTOREV after the release is done. I am thinking of documenting our release process (something like this: https://optee.readthedocs.io/en/latest/general/releases.html#release-procedure)

vj-kumar:
> > > > Can we conditionally build stable? Maybe defaulting to autorev but an option to build stable?
> > > 
> > > 
> > > I guess we could but was worried that the end-result would be a mix of stable and devel (since HEAD may contain changes to the image that aren't compatible with "stable"). I would recommend people wanting to build stable to checkout a specific tag (e.g. v0.15) so they get everything aligned. Thoughts?
> > 
> > 
> > Then again, in the event of release you would modify the changelog, add a fixed commit, tag it and then reintroduce a development version like now(Autorev + dev changelog)?
> 
> Yes that's right. We would use a fixed SRCREV when preparing our release (-rcX tags) and switch back to AUTOREV after the release is done. I am thinking of documenting our release process (something like this: https://optee.readthedocs.io/en/latest/general/releases.html#release-procedure)

We can probably introduce a release script later, so that we don't accidentaly forget. If the whole release itself can automatically done by CI that would be good too.

vj-kumar (approved):

PR #195, opened 2022-02-04
Use Debian kernel for the beaglebone-black

chombourger:
The Debian kernel appears to work - at least Ethernet, USB and USB Function are.

I however suspect that GPIO lines are numbered differently
I could use some help to determine what is going on


vj-kumar:
Triggering a local build to reproduce the issue.

jan-kiszka:
While reading the code, I was wondering already what motivated to select a downstream kernel for such an old SOC that works fine with mainline since ages (related commit log was empty...). Even better when that's also the case with the Debian config!

chombourger:
> Triggering a local build to reproduce the issue.

Used gpio28 (P9_12) as you suggested and the relay is correctly controlled

chombourger:
> While reading the code, I was wondering already what motivated to select a downstream kernel for such an old SOC that works fine with mainline since ages (related commit log was empty...). Even better when that's also the case with the Debian config!

we had another layer use that kernel. we started with a simple port of what we already have but had the goal to switch to mainline

.github/workflows/main.yml
```diff
@@ -63,6 +63,25 @@ jobs:
...
+        run: |
+          export KAS_BUILD_DIR=${PWD}/build-bbb
+          ./kas-container build kas/debian/mtda-beaglebone-black.yml
+      - name: Upload image
+        uses: actions/upload-artifact@v2
+        if: github.ref == 'refs/heads/master'
+        with:
+          name: mtda-nanopi-images
+          path: |
+            build-bbb/tmp/deploy/images/beaglebone-*/isar-image-base-mtda-*-beaglebone-*.wic.img
```
vj-kumar, line 83: We migrated to mtda-image recently. We would need to change this accordingly

.github/workflows/main.yml
```diff
@@ -63,6 +63,25 @@ jobs:
...
+        run: |
+          export KAS_BUILD_DIR=${PWD}/build-bbb
+          ./kas-container build kas/debian/mtda-beaglebone-black.yml
+      - name: Upload image
+        uses: actions/upload-artifact@v2
+        if: github.ref == 'refs/heads/master'
+        with:
+          name: mtda-nanopi-images
+          path: |
+            build-bbb/tmp/deploy/images/beaglebone-*/mtda-image-base-mtda-*-beaglebone-*.wic.img
```
vj-kumar, line 83: We name it just mtda-image.  Since we had only one variant  I did not include the -base suffix.

.github/workflows/main.yml
chombourger: fixed

vj-kumar (approved):

PR #197, opened 2022-02-04
power/docker: make sure we do not pull all images

chombourger:
Docker's images.pull() defaults to all_tags=False. There are
however reports that MTDA would fetch all images on start-up.
Call pull() with all_tags explicitly set to False.


chombourger:
Reference: https://docker-py.readthedocs.io/en/stable/images.html

vj-kumar (approved):

PR #200, opened 2022-02-07
setup: added script to auto install mtda aritfacts to system

boksheos:
As an end user perspective this script will help to install mtda and its artifacts automatically following are the scripts features
1) identify the distro  debian OR ubuntu if others gives an message unsupported and untested 
2) It will prompt user 5 options 1) mtda 2) mtda-kvm 3) mtda-docker 4) all 5) quit 
3) script will install all the above options 

Scope for improvement :
1) script can be extended to any future host package dependencies 
2) Can be extended to install LAVA server and respective setup 
3) future installation can be implemented for other distro support using pip

Note :  This script is initial draft  for mtda auto install which will help user in quick installation.
Signed-off-by: Onkar Bokshe <OnkarSurendrakumar_Bokshe@mentor.com>

### DIFF
--- a/scripts/setup-mtda
+++ b/scripts/setup-mtda
@@ -1,0 +1,175 @@
+#!/bin/bash
+# ---------------------------------------------------------------------------------------------------------------------
+# Setup Multi-Tenant Device Access on system
+# ---------------------------------------------------------------------------------------------------------------------
+#
+# This software is a part of MTDA.
+# Copyright (C) 2022 Siemens Digital Industries Software
+#
+
+# ---------------------------------------------------------------------------------------------------------------------
+# This script shall be run with admin privileges
+# ---------------------------------------------------------------------------------------------------------------------
+
+uid=$(id -u)
+if [ ${uid} -ne 0 ]
+then
+   echo "error: please run this script as root OR using sudo!" >&2
+   exit 1
+fi
+
+
+# ---------------------------------------------------------------------------------------------------------------------
+# Check distro of running system
+# ---------------------------------------------------------------------------------------------------------------------
+
+dist_name_version() {
+	if [ -f /etc/os-release ]; then
+		DIST=$(cat /etc/os-release | grep ^ID= |  cut -d"=" -f2)
+		# convert dist to lower case
+		DIST=$(echo ${DIST} | tr '[:upper:]' '[:lower:]')
+		case "${DIST}" in
+			rpb*) DIST="oe-rpb" ;;
+		esac
+		if [ "debian" = "$DIST" ]; then
+			DIST_VERSION=$(cat /etc/${DIST}_version)
+		else
+			DIST_VERSION=$(cat /etc/os-release | grep ^VERSION= | awk '{ print $1}' | cut -d'"' -f2)
+		fi
+	fi
+}
+
+# ---------------------------------------------------------------------------------------------------------------------
+# Check supported distro and its version
+# ---------------------------------------------------------------------------------------------------------------------
+dist_name_version
+if [ "debian" = "$DIST" ] && [ $(echo "$DIST_VERSION 10.10" | awk '{print ($1 >= $2)}') ]; then
+	echo "$DIST:$DIST_VERSION is supported distro for MTDA"
+elif [ "ubuntu" = "$DIST" ] && [ $(echo "$DIST_VERSION 20.05" | awk '{print ($1 >= $2)}') ]; then
+	echo "$DIST:$DIST_VERSION is supported distro for MTDA"
+else
+	echo "error:Unsupported distro: MTDA is not yet tested on this distro"
+	exit 1
+fi
+
+# ---------------------------------------------------------------------------------------------------------------------
+# Add apt mirror for MTDA 
+# ---------------------------------------------------------------------------------------------------------------------
+
+add_apt_mirror() {
+    case "${DIST}" in
+        debian)
+			echo 'deb [trusted=yes] https://apt.fury.io/mtda/ /' | \
+				tee /etc/apt/sources.list.d/mtda.list
+        ;;
+        ubuntu)
+			add-apt-repository ppa:chombourger/mtda-focal
+        ;;
+    esac
+}
+
+# ---------------------------------------------------------------------------------------------------------------------
+# MTDA and mtda-pytest installation 
+# ---------------------------------------------------------------------------------------------------------------------
+
+mtda_install() {
+	apt-get update
+	apt-get -y install mtda mtda-pytest
+}
+
+# ---------------------------------------------------------------------------------------------------------------------
+# docker installation 
+# ---------------------------------------------------------------------------------------------------------------------
+
+docker_install() {
+	apt-get -y remove docker docker-engine docker.io containerd runc
+	apt-get update
+	apt-get -y install \
+		ca-certificates \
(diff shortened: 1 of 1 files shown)


scripts/setup-mtda
```diff
@@ -0,0 +1,156 @@
...
+# ---------------------------------------------------------------------------------------------------------------------
+
+# ---------------------------------------------------------------------------------------------------------------------
+# Check distro of running system
+# ---------------------------------------------------------------------------------------------------------------------
+
+dist_name() {
+    if [ -f /etc/os-release ]; then
+        # shellcheck disable=SC1091
+        dist=$(. /etc/os-release && echo "${ID}")
```
chombourger, line 14: it is not recommended to source /etc/os-release

scripts/setup-mtda
```diff
@@ -0,0 +1,156 @@
...
+    elif [ -x /usr/bin/lsb_release ]; then
+        dist="$(lsb_release -si)"
+    elif [ -f /etc/lsb-release ]; then
+        # shellcheck disable=SC1091
+        dist="$(. /etc/lsb-release && echo "${DISTRIB_ID}")"
+    elif [ -f /etc/debian_version ]; then
+        dist="debian"
+    elif [ -f /etc/fedora-release ]; then
+        dist="fedora"
+		warn_msg "Unsupported distro: mtda installation is not yet tested on this system"
```
chombourger, line 24: we don't seem to have warn_msg defined anywhere
what is the source of this script?

scripts/setup-mtda
```diff
@@ -0,0 +1,156 @@
...
+	curl -fsSL https://download.docker.com/linux/debian/gpg | sudo gpg --dearmor -o /usr/share/keyrings/docker-archive-keyring.gpg
+	echo \
+	"deb [arch=$(dpkg --print-architecture) signed-by=/usr/share/keyrings/docker-archive-keyring.gpg] https://download.docker.com/linux/debian \
+		$(lsb_release -cs) stable" | sudo tee /etc/apt/sources.list.d/docker.list > /dev/null
+	apt-get update -y
+	apt-get -y install docker-ce docker-ce-cli containerd.io  
+	/sbin/adduser $USER docker
+	systemctl enable docker
+	systemctl start docker
+}
```
chombourger, line 100: Debian and Ubuntu provide docker.io

scripts/setup-mtda
```diff
@@ -0,0 +1,156 @@
+#!/bin/bash
+# ---------------------------------------------------------------------------------------------------------------------
+# Setup Multi-Tenant Device Access on system
+# Copyright (c) 2021-2022, Mentor Graphics, a Siemens business
```
vj-kumar, line 4: Copyright (C) 2022 Siemens Digital Industries Software

scripts/setup-mtda
```diff
@@ -0,0 +1,156 @@
...
+# ---------------------------------------------------------------------------------------------------------------------
+
+# ---------------------------------------------------------------------------------------------------------------------
+# Check distro of running system
+# ---------------------------------------------------------------------------------------------------------------------
+
+dist_name() {
+    if [ -f /etc/os-release ]; then
```
vj-kumar, line 12: IMO these checks might be enough to determine if we are in a debian based system
if /etc/debian_verison present -> Its debian
cat /etc/os-release | grep ^ID=   to know if ubuntu / debian
Version can be fetched like this, though, it could be empty. VERSION_ID is an optional field
cat /etc/os-release | grep ^VERSION_ID  to know if tested version, 20.04/10.x etc.

scripts/setup-mtda
boksheos: true i have pulled it from lava test suite updated it with echo "W:<>"

scripts/setup-mtda
```diff
@@ -0,0 +1,156 @@
...
+        rpb*) dist="oe-rpb" ;;
+    esac
+}
+
+# ---------------------------------------------------------------------------------------------------------------------
+# This script shall be run with admin privileges
+# ---------------------------------------------------------------------------------------------------------------------
+
+uid=$(id -u)
+if [ ${uid} -ne 0 -o -z "${SUDO_USER}" ]
```
vj-kumar, line 47: Is "id -u" not sufficient? As I understand from manpage SUDO_USER, is set to the user that invoked sudo, that means that the id -u is already 0 and the first condition is met?

scripts/setup-mtda
chombourger: what's the copyright/license of that script?

scripts/setup-mtda
```diff
@@ -0,0 +1,156 @@
...
+	apt-get update -y
+	apt-get  -y install mtda mtda-pytest
+}
+
+# ---------------------------------------------------------------------------------------------------------------------
+# docker installation 
+# ---------------------------------------------------------------------------------------------------------------------
+
+docker_install() {
+	apt-get -y remove docker docker-engine docker.io containerd runc
```
vj-kumar, line 84: Not sure why we need to remove those?

scripts/setup-mtda
```diff
@@ -0,0 +1,156 @@
...
+   exit 1
+fi
+
+# ---------------------------------------------------------------------------------------------------------------------
+# Add apt mirror for MTDA 
+# ---------------------------------------------------------------------------------------------------------------------
+
+add_apt_mirror() {
+    dist_name
+    case "${dist}" in
```
vj-kumar, line 59: DIST instead of dist maybe?

scripts/setup-mtda
vj-kumar: This could just be outside in the main script and you can take a call based on distribution whether to proceed running the script or not.

scripts/setup-mtda
```diff
@@ -0,0 +1,156 @@
...
+    esac
+}
+
+# ---------------------------------------------------------------------------------------------------------------------
+# MTDA and mtda-pytest installation 
+# ---------------------------------------------------------------------------------------------------------------------
+
+mtda_install() {
+	apt-get update -y
+	apt-get  -y install mtda mtda-pytest
```
vj-kumar, line 76: Extra space after apt-get

scripts/setup-mtda
```diff
@@ -0,0 +1,156 @@
...
+	apt-get  -y install \
+		ca-certificates \
+		curl \
+		gnupg \
+		lsb-release
+	curl -fsSL https://download.docker.com/linux/debian/gpg | sudo gpg --dearmor -o /usr/share/keyrings/docker-archive-keyring.gpg
+	echo \
+	"deb [arch=$(dpkg --print-architecture) signed-by=/usr/share/keyrings/docker-archive-keyring.gpg] https://download.docker.com/linux/debian \
+		$(lsb_release -cs) stable" | sudo tee /etc/apt/sources.list.d/docker.list > /dev/null
+	apt-get update -y
```
vj-kumar, line 95: I don't think update ever asks for user input. We could drop the -y.

scripts/setup-mtda
vj-kumar: @chombourger  Or is it fine?

scripts/setup-mtda
vj-kumar: So you want to avoid running this as root user? Are there any reasons to avoid root user?

PR #201, opened 2022-02-08
meta-isar: remove name resolution message during sudo

bhargavdas:
Signed-off-by: Bhargav Das <bhargav_das@mentor.com>

### DIFF
--- a/meta-isar/recipes-conf/hostname/files/postinst
+++ b/meta-isar/recipes-conf/hostname/files/postinst
@@ -3,3 +3,6 @@
 set -e
 
 echo "mtda" > /etc/hostname
+
+# add hostname to /etc/hosts
+echo "127.0.1.1   mtda" >> /etc/hosts
--- a/meta-isar/recipes-conf/hostname/mtda-hostname_0.4.bb
+++ b/meta-isar/recipes-conf/hostname/mtda-hostname_0.4.bb
@@ -4,7 +4,7 @@
 
 DESCRIPTION = "set system hostname to mtda"
 MAINTAINER = "Cedric Hombourger <chombourger@gmail.com>"
-DEBIAN_DEPENDS = "base-files"
+DEBIAN_DEPENDS = "base-files, netbase"
 DPKG_ARCH = "all"
 
 SRC_URI = "file://postinst"


meta-isar/recipes-conf/hostname/files/postinst
```diff
@@ -3,3 +3,8 @@
 set -e
 
 echo "mtda" > /etc/hostname
+echo "127.0.0.1    localhost\n/
+    127.0.1.1    mtda\n/
+    ::1    localhost ip6-localhost ip6-loopback\n/
+    ff02::1    ip6-allnodes\n/
+    ff02::2    ip6-allrouters" >> /etc/hosts
```
chombourger, line 10: can you check if adding the `netbase` package to the image would help?
see /var/lib/dpkg/info/netbase.postinst - it is supposed to create /etc/hosts

meta-isar/recipes-conf/hostname/files/postinst
```diff
@@ -3,3 +3,8 @@
 set -e
 
 echo "mtda" > /etc/hostname
+echo "127.0.0.1    localhost\n/
```
vj-kumar, line 6: I am not sure why you need \n/ at the end.

I think the below should be sufficient.
echo "127.0.0.1    localhost
127.0.1.1    mtda
::1    localhost ip6-localhost ip6-loopback
ff02::1    ip6-allnodes
ff02::2    ip6-allrouters" >> /etc/host

meta-isar/recipes-conf/hostname/files/postinst
```diff
@@ -3,3 +3,10 @@
 set -e
 
 echo "mtda" > /etc/hostname
+
+# add hostname in /etc/hosts
+echo "127.0.0.1    localhost
+    127.0.1.1    mtda
```
chombourger, line 9: this is the only line we need to add

chombourger:
add DEBIAN_DEPENDS = "netbase" to the recipe (`.bb` file)

meta-isar/recipes-conf/hostname/files/postinst
bhargavdas: I tried adding only `echo "127.0.1.1    mtda" >> /etc/hosts` but it overrides the other contents the file. That's I added all the lines

vj-kumar: >> should basically append. Just tried it to confirm.

bhargavdas:
> add DEBIAN_DEPENDS = "netbase" to the recipe (`.bb` file)

getting the following error 
`dpkg-gencontrol: warning: can't parse dependency base-files netbase`
after adding 
```
DEBIAN_DEPENDS = "base-files netbase"
```

chombourger:
> > add DEBIAN_DEPENDS = "netbase" to the recipe (`.bb` file)
> 
> getting the following error `dpkg-gencontrol: warning: can't parse dependency base-files netbase` after adding
> 
> ```
> DEBIAN_DEPENDS = "base-files netbase"
> ```
DEBIAN_DEPENDS = "base-files, netbase"

(entries are comma-separated in debian/control)

meta-isar/recipes-conf/hostname/files/postinst
chombourger: > I tried adding only `echo "127.0.1.1 mtda" >> /etc/hosts` but it overrides the other contents the file. That's why I added all the lines

even with DEBIAN_DEPENDS correctly set?

bhargavdas: that I have not checked.checking now

bhargavdas: it is appending after adding netbase package.

meta-isar/recipes-conf/hostname/files/postinst
```diff
@@ -3,3 +3,6 @@
 set -e
 
 echo "mtda" > /etc/hostname
+
+# add hostname in /etc/hosts
```
chombourger, line 7: add hostname to /etc/hosts

chombourger:
let's squash the 2 commits (git rebase -i origin/master and use `squash` instead of `pick` for the second commit)
let's rename the commit to: `meta-isar/recipes-conf: add "mtda" to /etc/hosts`

chombourger (approved):

chombourger:
Thanks @bhargavdas for your contribution!

PR #206, opened 2022-02-09
ci: auto assign reviewer on pull_request_target

ricbha:


ricbha:
@chombourger Added the change but still getting the same warning. I am not sure if this will work but we can try.

chombourger (approved):

PR #212, opened 2022-02-14
add video and virtual-keyboard to the web-based UI

chombourger:


vj-kumar (approved):
I don't understand Javascript yet. Skipped those parts.

Issue #210, opened 2022-02-14
stop using isar-image-base or at least set an own image DESCRIPTION


chombourger:
created from #209 as suggested by @jan-kiszka

PR #216, opened 2022-02-22
meta-isar: create our own target image

vj-kumar:
Create our own variation of image called mtda-image.

Closes: #210

Signed-off-by: Vijai Kumar K <Vijaikumar_Kanagarajan@mentor.com>

### DIFF
--- a/meta-isar/recipes-core/images/mtda-image.bb
+++ b/meta-isar/recipes-core/images/mtda-image.bb
@@ -5,6 +5,15 @@
 # SPDX-License-Identifier: MIT
 # ---------------------------------------------------------------------------
 
+inherit image
+
+DESCRIPTION = "Debian image for MTDA assist boards"
+
+LICENSE = "MIT"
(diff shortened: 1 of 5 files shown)


meta-isar/recipes-core/images/mtda-image.bb
```diff
@@ -5,6 +5,15 @@
 # SPDX-License-Identifier: MIT
 # ---------------------------------------------------------------------------
 
+inherit image
+
+DESCRIPTION = "Debian image for MTDA assist boards"
+
+LICENSE = "gpl-2.0"
+LIC_FILES_CHKSUM = "file://${LAYERDIR_core}/licenses/COPYING.GPLv2;md5=751419260aa954499f7abaabaa882bbe"
```
chombourger, line 13: MIT to be consistent with the header of this recipe?

vj-kumar: We could use MIT.

chombourger (approved):
Looks good and CI checks passed. Merging

PR #229, opened 2022-03-06
console/logger: Open console unconditionally

jan-kiszka:
The there is no power controller, assume that the device is already
powered. Without this, the console does not work in most cases.

Keeping the suspicious locking pattern untouched for now, that one is
likely broken as well.

### DIFF
--- a/mtda/console/logger.py
+++ b/mtda/console/logger.py
@@ -303,8 +303,9 @@ class ConsoleLogger:
 
         if self.power_controller is not None:
             self.power_controller.wait()
-            with self.rx_lock:
-                con.open()
+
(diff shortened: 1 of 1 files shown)


chombourger:
Should commit log start with: *When* there is no power controller, ...

(Replace "the" with "when"?)

jan-kiszka:
Hmpf, I'm seeing problems again, even with this applied and the service started manually. More investigation needed.

jan-kiszka:
OK, turning on `debug=4` makes it work again. Looks like my nose was right regarding "locking is broken, races ahead!"

chombourger (approved):

PR #231, opened 2022-03-06
meta-isar/recipes-python: Build local version

jan-kiszka:
This avoids pulling random versions from github, permitting to package
local changes as well as clearly defined releases. The price is the need
of maintaining a list of top-level files and directories.

As the resulting input for the package build is now a local folder, the
source/format used for the Isar build needs to be switched to default.

### DIFF
--- a/meta-isar/recipes-python/mtda/mtda_git.bb
+++ b/meta-isar/recipes-python/mtda/mtda_git.bb
@@ -7,8 +7,37 @@
 
 inherit dpkg
 
-SRC_URI += "git://git@github.com/siemens/mtda.git;protocol=https;branch=master"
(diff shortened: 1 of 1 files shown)


jan-kiszka:
This one still has some issue, apparently packaging stall files. Give me a moment.

chombourger (approved):

chombourger:
just noting that during development, we usually keep the same image, build a local version of the mtda package using `dpkg-buildpackage` and updating the generated package on the assist board (such as the NanoPI)

jan-kiszka:
I'm not on Debian, so I'm also benefiting from `kas-container build ... --target mtda`.

PR #233, opened 2022-03-06
Misc fixes

chombourger:
This is a subset of #226 (observing sporadic hangs in `_stop_composite`

PR #234, opened 2022-03-06
power: Add shellcmd driver

jan-kiszka:
This driver allows to manage power switches via simple shell commands.
It is a break-out box for a broad range of bindings, e.g. by using curl
to talk to the REST interface of a power control.

### DIFF
--- a/docs/config.rst
+++ b/docs/config.rst
@@ -80,7 +80,7 @@ General settings
 
   * ``variant``: string [required]
       Select a power variant from ``aviosys_8800``, ``gpio``, ``pduclient``, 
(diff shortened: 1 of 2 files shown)


jan-kiszka:
Will provide a rebase over #233 once it's ready - makes things slimmer.

chombourger (approved):
it is a brilliant idea, it should be easy to integrate with many smart PDUs that are out there

jan-kiszka:
Yeah, I'm using it here with Tasmota-enabled smart sockets. Was briefly considering to implement a MQTT client instead, but that turned out out be more complex with no added value.

PR #236, opened 2022-03-06
meta-isar: Add Raspberry Pi 4 model B

jan-kiszka:
This still needs some documentation, but I wanted to get the integration out first. It now serves as a local agent for booting a IOT2050 from USB and also remotely flashing its firmware (extra local recipe for sf100linux).

My local device only has 1G RAM. However, with some (currently out-of-stock) 8G RPi4, it should be possible to run the USB storage completely out of tmpfs, avoiding writes to persistent storage. Likely interesting for CI.


chombourger:
> This still needs some documentation, but I wanted to get the integration out first. It now serves as a local agent for booting a IOT2050 from USB and also remotely flashing its firmware (extra local recipe for sf100linux).
> 
> My local device only has 1G RAM. However, with some (currently out-of-stock) 8G RPi4, it should be possible to run the USB storage completely out of tmpfs, avoiding writes to persistent storage. Likely interesting for CI.

I like the idea of using tmpfs for our USB mass storage function

jan-kiszka:
I just played with io2050-base-image (~400MB), and that worked like charm. We can get read/write rates of almost 40 MB/s on the target this way!

chombourger:
```
# MISSING COPYRIGHT AND LICENSING INFORMATION

The following files have no copyright and licensing information:
* meta-isar/recipes-bsp/rpi-firmware/files/debian/cmdline.txt
* meta-isar/recipes-bsp/rpi-firmware/files/debian/config.txt
* meta-isar/recipes-bsp/rpi-firmware/files/debian/rpi-firmware.install
* meta-isar/recipes-bsp/rpi-firmware/files/rules
```

jan-kiszka:
Will fix what can be fixed. How to exclude files from this check? I suspect that at least cmdline.txt does not support comments.

chombourger:
> Will fix what can be fixed. How to exclude files from this check? I suspect that at least cmdline.txt does not support comments.

see .reuse/dep5

PR #226, opened 2022-03-04
runtime switching of the USB Mass Storage function

chombourger:

PR #239, opened 2022-04-17
storage/usbf: Remove and re-install storage when switching between host and target

jan-kiszka:
This allows to replace the backing file while it is in host mode and
informs the USB gadget about changes to it. Also, the target will not be
able to access it anymore while it is in host mode.

Maybe it is possible to further automate updates by unconditionally reinstalling the device on target power-on.

### DIFF
--- a/mtda/storage/usbf.py
+++ b/mtda/storage/usbf.py
@@ -94,6 +94,8 @@ class UsbFunctionController(Image):
         self.mtda.debug(3, "storage.usbf.to_host()")
         self.lock.acquire()
 
+        Composite.remove()
+
         self.mode = self.SD_ON_HOST
         result = True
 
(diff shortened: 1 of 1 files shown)


chombourger:
thank you @jan-kiszka 
the question is whether installing/removing the gadget is working reliably?
I have observed sporadic hangs when I worked on #226

chombourger:
> Maybe it is possible to further automate updates by unconditionally reinstalling the device on target power-on

we would probably want to make this a config option. We have some x86 devices that are unconditionally booting from USB if detected. we wanted to use this ability to turn the gadget ON/OFF to either boot from internal or external storage

jan-kiszka:
At least on the RPi4 where I'm using this now, I'm not seeing any problems. I switched there from `mtda-cli storage write` to NFS-backed images, streamed from the builder. That saves quite some deployment time - would save even more if fs-cache would work in combination with in-kernel usage by the usb function.

Regarding your use case of keeping the gadget away from the target only for booting: That would definitely require run-time switching of the ownership and, thus, the attached/detached state as this patch manages. So far, you can only decide to unplug it or re-plug it while the target is off.

chombourger (approved):

chombourger:
> At least on the RPi4 where I'm using this now, I'm not seeing any problems. I switched there from `mtda-cli storage write` to NFS-backed images, streamed from the builder. That saves quite some deployment time - would save even more if fs-cache would work in combination with in-kernel usage by the usb function.
> 
> Regarding your use case of keeping the gadget away from the target only for booting: That would definitely require run-time switching of the ownership and, thus, the attached/detached state as this patch manages. So far, you can only decide to unplug it or re-plug it while the target is off.

ok. that sounds good. will merge this and recheck on my end with your changes

Issue #242, opened 2022-05-19
zerorpc 0.6.3 is not compatible with pyzmq 23.0.0


chombourger:
significant changes were introduced in pyzmq 23.0.0

our tests are failing with
`AttributeError: module 'zerorpc.gevent_zmq' has no attribute 'PUSH'`

need to work with the zerorpc project to fix such errors.
check if we can modify our requirements so that we don't run with pyzmp >= 23.0.0 for now

chombourger:
submitted https://github.com/0rpc/zerorpc-python/pull/252

PR #243, opened 2022-05-19
setup: require pyzmq<23.0.0

chombourger:
zerorpc fails with errors such as 'zerorpc.gevent_zmq' has no attribute
'PUSH' when running against the latest version. Changes to zerorpc are
probably required.

Closes: #242

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -52,7 +52,7 @@ and testers to remotely access and control hardware devices.
         "pyserial>=2.6",
         "python-daemon>=2.0",
         "pyusb>=1.0",
-        "pyzmq>=15.0",
+        "pyzmq>=15.0,<23.0.0",
         "psutil",
         "requests",
         "zerorpc>=0.6.0",


chombourger:
This PR fixes our CI

vj-kumar:
It is to note that when these are converted to debian dependencies the version requirements are not captured per se. 
Maybe we should add this corresponding debian package and version check in the control file as well.

setup.py
```diff
@@ -52,7 +52,7 @@
         "pyserial>=2.6",
         "python-daemon>=2.0",
         "pyusb>=1.0",
-        "pyzmq>=15.0",
+        "pyzmq>=15.0,<23.0.0",
```
vj-kumar, line 55: It is to note that when these are converted to debian dependencies the version requirements are not captured per se. 
Maybe we should add this corresponding debian package and version check in the control file as well.

chombourger: My thinking was to add the zerorpc patch i submitted upstream to our Debian build

vj-kumar: We could do that as well.

PR #253, opened 2022-12-01
scripts: add support for loading device scripts

chombourger:
Ease sharing of device scripts that are used to boot from USB, reset the TPM, enable Secure Boot, etc. by having MTDA load them from mtda.script.<variant> with <variant> being loaded from the [environment] section of the configuration file. Sample scripts for the Siemens SIMATIC IPC227E are provided with this changeset.

### DIFF
--- a/configs/qemu.ini
+++ b/configs/qemu.ini
@@ -16,6 +16,12 @@
 debug = 2
 
 # ---------------------------------------------------------------------------
+# Initial environment
+# ---------------------------------------------------------------------------
+[environment]
+variant=qemu
+
+# ---------------------------------------------------------------------------
 # Remote settings
 # ---------------------------------------------------------------------------
 [remote]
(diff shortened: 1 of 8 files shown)


vj-kumar:
If a user writes a custom script for their target, where should they put them?

chombourger:
> If a user writes a custom script for their target, where should they put them?

For now `/usr/lib/python3/dist-packages/mtda/scripts/` but will likely want a mechanism to let the user specify/use arbitrary paths

vj-kumar:
> 

IMO, That could be in the environment section as well. Something like below. Multiple paths could be seperated using ":" or something equivalent. 
```
[environment]
variant=qemu
paths="/opt/custom-scripts:/etc/mtda/custom-scripts"
```

chombourger:
> > 
> 
> IMO, That could be in the environment section as well. Something like below. Multiple paths could be seperated using ":" or something equivalent.
> 
> ```
> [environment]
> variant=qemu
> paths="/opt/custom-scripts:/etc/mtda/custom-scripts"
> ```

yes `paths=` or maybe `scripts=`
e.g. in the future we could add a library of reference images to let users install "My OS v1.1" over MTDA and then attempt an upgrade to the version they are testing. for that we could use `images=`

vj-kumar (approved):

PR #266, opened 2023-01-09
console/serial: remove check for device from configure_systemd()

chombourger:


chombourger:
does it look OK @vj-kumar ?

vj-kumar (approved):

PR #268, opened 2023-01-16
docs/build: remove .img from ISAR image name

bovi:
ISAR generates an image without .img extension.


vj-kumar (approved):

chombourger:
thank you so much @bovi for your feedback and pull request

chombourger (approved):

PR #273, opened 2023-01-18
mtda-cli: prohibit user from daemonizing client

shikl3x7:
mtda-cli may be either used in client or server mode but not
both. Reject use of -d and -r options together.

Signed-off-by: Shivaschandra KL <shivaschandra.k-l@siemens.com>

### DIFF
--- a/mtda-cli
+++ b/mtda-cli
@@ -35,7 +35,7 @@ class Application:
 
     def __init__(self):
         self.agent = None
-        self.remote = "localhost"
+        self.remote = None
         self.logfile = "/var/log/mtda.log"
         self.pidfile = "/var/run/mtda.pid"
         self.exiting = False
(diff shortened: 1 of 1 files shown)


jan-kiszka:
Calling the daemon "...-cli" is already an interface inconsistency. You should have a different command for starting the daemon, even if that should internally map to the same python script.

chombourger:
> Calling the daemon "...-cli" is already an interface inconsistency. You should have a different command for starting the daemon, even if that should internally map to the same python script.

this is indeed in the list of things to do

chombourger (approved):

vj-kumar (approved):

chombourger:
commit log should read something like:
```
mtda-cli: prohibit user from daemonizing client

mtda-cli may be either used in client or server mode but not
both. Reject use of -d and -r options together.

Signed-off-by: Shivaschandra KL <shivaschandra.k-l@siemens.com>
```

Issue #116, opened 2021-12-22
refactor mtda-cli: create separate applications for our daemon and client


chombourger:
The `mtda-cli` script handles both running mtda as daemon as well as a client. Since there isn't much in common between these two modes, it may be desirable to separate them out (at the source level and package level)

PR #278, opened 2023-02-01
mtda-service: move service code from mtda-cli to its own module

chombourger:
Closes: #116

### DIFF
--- a/mtda/main.py
+++ b/mtda/main.py
@@ -111,11 +111,6 @@ class MultiTenantDeviceAccess:
         self.version = __version__
         self._www = None
 
-        # Config file in $HOME/.mtda/config
-        home = os.getenv('HOME', '')
-        if home != '':
-            self.config_files.append(os.path.join(home, '.mtda', 'config'))
-
         # Config file in /etc/mtda/config
         if os.path.exists('/etc/mtda'):
             self.config_files.append(os.path.join('/etc', 'mtda', 'config'))
@@ -125,6 +120,11 @@ class MultiTenantDeviceAccess:
             fragments = glob.glob('/etc/mtda/config.d/*.conf')
             if len(fragments) > 0:
                 self.config_files.extend(sorted(fragments))
+
+        # Config file in $HOME/.mtda/config
+        home = os.getenv('HOME', '')
+        if home != '':
+            self.config_files.append(os.path.join(home, '.mtda', 'config'))
 
     def agent_version(self):
         return self.version
--- a/mtda/main.py
+++ b/mtda/main.py
@@ -117,8 +117,14 @@ class MultiTenantDeviceAccess:
             self.config_files.append(os.path.join(home, '.mtda', 'config'))
 
         # Config file in /etc/mtda/config
-        if os.path.exists('/etc'):
+        if os.path.exists('/etc/mtda'):
             self.config_files.append(os.path.join('/etc', 'mtda', 'config'))
+
+        # Config fragments in /etc/mtda/config.d/
+        if os.path.exists('/etc/mtda/config.d'):
+            fragments = glob.glob('/etc/mtda/config.d/*.conf')
+            if len(fragments) > 0:
+                self.config_files.extend(fragments)
 
     def agent_version(self):
         return self.version
--- a/mtda/main.py
+++ b/mtda/main.py
@@ -124,7 +124,7 @@ class MultiTenantDeviceAccess:
         if os.path.exists('/etc/mtda/config.d'):
             fragments = glob.glob('/etc/mtda/config.d/*.conf')
             if len(fragments) > 0:
-                self.config_files.extend(fragments)
+                self.config_files.extend(sorted(fragments))
 
     def agent_version(self):
         return self.version
--- a/configs/www.conf
+++ b/configs/www.conf
@@ -1,0 +1,15 @@
+# ---------------------------------------------------------------------------
+# www configuration fragment for MTDA
+# ---------------------------------------------------------------------------
+#
+# This software is a part of MTDA.
+# Copyright (C) 2023 Siemens Digital Industries Software
+#
+# ---------------------------------------------------------------------------
+# SPDX-License-Identifier: MIT
+# ---------------------------------------------------------------------------
+
+[www]
+port=9080
+# Uncomment the following to enable connections from anywhere
+# host=0.0.0.0
--- a/debian/mtda.service
+++ b/debian/mtda.service
@@ -1,5 +1,7 @@
 [Unit]
 Description=Multi-Tenant Device Access
+After=network-online.target
+Wants=network-online.target
 
 [Service]
 Type=simple
--- a/debian/mtda.service
+++ b/debian/mtda.service
@@ -3,7 +3,7 @@ Description=Multi-Tenant Device Access
 
 [Service]
 Type=simple
-ExecStart=/usr/bin/mtda-cli -d -n
+ExecStart=/usr/sbin/mtda-service -n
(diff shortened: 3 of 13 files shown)


bhargavdas:
Did a quick test by rebasing these changes (and making changes to the yocto layers) with yocto qemuarm machine, everything seems to be working on a high level.
Except when the system starts the `mtda.service` is in failed state. I think this might due to missing config file.

chombourger:
> Did a quick test by rebasing these changes (and making changes to the yocto layers) with yocto qemuarm machine, everything seems to be working on a high level. Except when the system starts the `mtda.service` is in failed state. I think this might due to missing config file.

Good to know. I assume we could add a conditional to the service to make it start only if /etc/mtda/config exists

mtda/main.py
```diff
@@ -117,9 +117,15 @@ def __init__(self):
             self.config_files.append(os.path.join(home, '.mtda', 'config'))
 
         # Config file in /etc/mtda/config
-        if os.path.exists('/etc'):
+        if os.path.exists('/etc/mtda'):
             self.config_files.append(os.path.join('/etc', 'mtda', 'config'))
 
+        # Config fragments in /etc/mtda/config.d/
+        if os.path.exists('/etc/mtda/config.d'):
+            fragments = glob.glob('/etc/mtda/config.d/*.conf')
```
vj-kumar, line 125: What will be the parse order for multiple configs. I see that glob.glob does not return a sorted list. So the traditional 00_test.conf  01_test2.conf would not work?

chombourger: good point. added a call to `sorted()`

vj-kumar: I am not sure if we are already following some kind of precedence. From the document I understand that we parse in this order 
```
 * ``$HOME/.mtda/config``
 * ``/etc/mtda/config``
 * ``/etc/mtda/config.d/*.conf``
```

Does that mean '/etc/mtda/config.d/*.conf' has higher precedence? Or the one in HOME. Can we update document to explain the precedence?

bhargavdas:
> > Did a quick test by rebasing these changes (and making changes to the yocto layers) with yocto qemuarm machine, everything seems to be working on a high level. Except when the system starts the `mtda.service` is in failed state. I think this might due to missing config file.
> 
> Good to know. I assume we could add a conditional to the service to make it start only if /etc/mtda/config exists

The condition to the service works, but still when the target (using yocto qemuarm here) boots/rebooted the `mtda.service` is in dead state initially and needs to started manually. Is this expected?
Here is the trimmed log,
1. target is booted first time without /etc/mtda/config
```
mtda login: root
root@mtda:~# systemctl list-units | grep mtda
  mtda-config.path                                                  loaded active     running         mtda-config.path
  mtda-config.service                                               loaded active     running         Service to monitor changes to mtda config file
  mtda.service                                                      loaded inactive   dead      start Multi-Tenant Device Access
root@mtda:~# systemctl start mtda
root@mtda:~# systemctl status mtda
* mtda.service - Multi-Tenant Device Access
     Loaded: loaded (/lib/systemd/system/mtda.service; enabled; vendor preset: enabled)
     Active: inactive (dead)
  Condition: start condition failed at Wed 2023-02-08 15:35:23 UTC; 24s ago
             `- ConditionPathExists=/etc/mtda/config was not met

Feb 08 15:35:23 mtda systemd[1]: Multi-Tenant Device Access was skipped because of a failed condition check (ConditionPathExists=/etc/mtda/config).
root@mtda:~# 
```
2. Created config
```
root@mtda:~#  cat /etc/mtda/config
[www]
port=9080
root@mtda:~# systemctl restart mtda
root@mtda:~# systemctl status mtda
* mtda.service - Multi-Tenant Device Access
     Loaded: loaded (/lib/systemd/system/mtda.service; enabled; vendor preset: enabled)
     Active: active (running) since Wed 2023-02-08 15:36:59 UTC; 1s ago
   Main PID: 257 (python3)
      Tasks: 1 (limit: 563)
     Memory: 5.9M
     CGroup: /system.slice/mtda.service
             `- 257 python3 /usr/sbin/mtda-service -n

Feb 08 15:36:59 mtda systemd[1]: Started Multi-Tenant Device Access.
root@mtda:~#
```
3. Rebooted the target
```
root@mtda:~# systemctl list-units | grep mtda
  mtda-config.path                                                  loaded active     running         mtda-config.path
  mtda-config.service                                               loaded active     running         Service to monitor changes to mtda config file
  mtda.service                                                      loaded inactive   dead      start Multi-Tenant Device Access
root@mtda:~# systemctl status mtda
* mtda.service - Multi-Tenant Device Access
     Loaded: loaded (/lib/systemd/system/mtda.service; enabled; vendor preset: enabled)
     Active: inactive (dead)
root@mtda:~#
root@mtda:~# cat /etc/mtda/config
[www]
port=9080
root@mtda:~#
```

bhargavdas:
> The condition to the service works, but still when the target (using yocto qemuarm here) boots/rebooted the `mtda.service` is in dead state initially and needs to started manually. Is this expected? I doubt if this behavior is only observable in yocto builds, will trigger a nanopi build and check

Checked in nanopi, this is only a yocto issue maybe.

chombourger:
> > The condition to the service works, but still when the target (using yocto qemuarm here) boots/rebooted the `mtda.service` is in dead state initially and needs to started manually. Is this expected? I doubt if this behavior is only observable in yocto builds, will trigger a nanopi build and check
> 
> Checked in nanopi, this is only a yocto issue maybe.

yes I believe our Isar images ship a mtda-config.service that cause systemd to restart mtda.service when /etc/mtda/config is changed

bhargavdas:
> yes I believe our Isar images ship a mtda-config.service that cause systemd to restart mtda.service when /etc/mtda/config is changed

Even the yocto image ships this file and the service starts prior to this PR. I wonder if something might be missing in systemd files.

bhargavdas:
> > yes I believe our Isar images ship a mtda-config.service that cause systemd to restart mtda.service when /etc/mtda/config is changed
> 
> Even the yocto image ships this file and the service starts prior to this PR. I wonder if something might be missing in systemd files.

Found it, the `Wants` section is blocking the start of the service.
```
Wants=network-online.target
```

bhargavdas:
> > > yes I believe our Isar images ship a mtda-config.service that cause systemd to restart mtda.service when /etc/mtda/config is changed
> > 
> > 
> > Even the yocto image ships this file and the service starts prior to this PR. I wonder if something might be missing in systemd files.
> 
> Found it, the `Wants` section is blocking the start of the service.
> 
> ```
> Wants=network-online.target
> ```

The `Wants` unit requires the service to be active, but in yocto's case `network-online.target` dies after acknowledging that the network is online. In debian (I have checked in ubuntu) it remains alive always. Can this be changed to maybe, `network.target` ? Else I will have patch the file during the yocto build.

bhargavdas:
> > > > yes I believe our Isar images ship a mtda-config.service that cause systemd to restart mtda.service when /etc/mtda/config is changed
> > > 
> > > 
> > > Even the yocto image ships this file and the service starts prior to this PR. I wonder if something might be missing in systemd files.
> > 
> > 
> > Found it, the `Wants` section is blocking the start of the service.
> > ```
> > Wants=network-online.target
> > ```
> 
> The `Wants` unit requires the service to be active, but in yocto's case (qemu) `network-online.target` dies after acknowledging that the network is online. In debian (I have checked in ubuntu) it remains alive always. Can this be changed to maybe, `network.target` ? Else I will have patch the file during the yocto build.

vj-kumar (approved):

PR #280, opened 2023-02-09
Fix flake8 warnings for mtda config

vj-kumar:


tox.ini
```diff
@@ -20,6 +20,7 @@ deps =
     reuse
     python-debian < 0.1.44
 commands =
+    flake8 mtda-config
```
chombourger, line 23: since a rebase is necessary, I would suggest adding `mtda-config` to the existing call to flake8 below

vj-kumar: Done.

chombourger (approved):

PR #287, opened 2023-02-20
mtda: use tell() to report how many writes were completed

chombourger:
Some decompressors may perform writes in the background and/or after the final call to write(). Add tell() to the storage interface to report the current position within the storage instead of calculating it from the some of return values from write(). This fixes the issue of no progress being shown after a zst stream was consumed.


chombourger:
@vj-kumar did you have a chance to check this pull request?

vj-kumar (approved):

PR #297, opened 2023-06-11
kas/debian: switch to bookworm for the nanopi-neo

chombourger:


kas/debian/bookworm-base.yml
```diff
@@ -0,0 +1,17 @@
+# ---------------------------------------------------------------------------
+# Base kas file for all configurations based on Debian bullseye
```
vj-kumar, line 2: Bookworm

chombourger: oops, had missed that one. fixed.

vj-kumar (approved):

PR #301, opened 2023-07-21
improve caching of mtda isar layer

fmoessbauer:
The layerdir variable contains an absolute path that is immediately expanded and points into the build host. This breaks sstate caching for tasks that reference this variable when building from a different path (like in the gitlab-ci). To fix this, we simply set the vardepvalue to a fixed value. As this variable just points to the current layer, there are no logical external changes expected and by that, this "pinning" is fine.


chombourger (approved):

chombourger:
Thank you @fmoessbauer for this PR. This is now merged.

PR #302, opened 2023-07-28
Add support to run MTDA web frontend behind reverse proxy

fmoessbauer:
This change ensures that all resources are references relative to the
current URI. By that, the whole MTDA web frontend can be served below a
subpath. This is required to support use-cases where multiple MTDA
devices are served by a single nginx reverse proxy in a single web page.
This approach can be used to serve a whole testing network via a single
IP entrypoint and without giving the users IP access to the testing
network itself.

To make the socket.io communication work, we further set the path
component of the current URI when connecting to the ws server.

### DIFF
--- a/.reuse/dep5
+++ b/.reuse/dep5
@@ -155,3 +155,6 @@ Files: mtda/assets/xterm.js
 Copyright: 2014 The xterm.js authors.
 License: MIT
 
+Files: mtda/share/nginx-multi-mtda.conf
+Copyright: 2023 Siemens AG
+License: MIT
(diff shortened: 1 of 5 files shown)


chombourger (approved):
Excellent work @fmoessbauer - thank you so much for your contribution

bovi:
@fmoessbauer I think the nginx config needs to be licensed. see here: https://github.com/siemens/mtda/blob/master/.reuse/dep5

fmoessbauer:
> @fmoessbauer I think the nginx config needs to be licensed. see here: https://github.com/siemens/mtda/blob/master/.reuse/dep5

Ok, will update the commit.

Issue #300, opened 2023-06-27
novnc bookworm package leads to the terminal of the mtda-www webview not working


kergos:
I built an image with the newly released debian bookworm image for the nanopi-neo.
After installing the mtda-www package, the terminal in the web-view is not loaded anymore.

This is because the current novnc bookworm package (https://packages.debian.org/bookworm/all/novnc/filelist) does not provide the required file `/usr/share/novnc/vendor/promise.js` (https://packages.debian.org/bullseye/all/novnc/filelist).

Apparently this was removed from novnc some time ago (https://github.com/novnc/noVNC/commit/890cff921d78835762fbdcfef862571c52cb3035)
This leads to the terminal in the web-view not properly loading. With the bullseye image it still works as intended.

chombourger:
Thanks for reporting @kergos

looks like `promise.js` gets included from `mtda/templates/index.html`
any idea how to fix this? should I look into this?

kergos:
If I understand it correctly  by looking at the basic example from novnc, (https://github.com/novnc/noVNC/blob/e8ad466e45a21598debe81824744d85ff2323c7b/vnc_lite.html) it seems like it could be changed. 

Basically `mtda/templates/index.html` would be needed to be rewritten somewhat to implement the new way to do it.
I currently lack the knowledge how to implement this properly and would need to read up on it, since it has been quite some time since I did anything web-related.

bovi:
@kergos I suggest we look at this and provide a merge request afterwards here. let's talk tomorrow.

kergos:
Fixed in #303

PR #303, opened 2023-08-01
Add mtda-www to image and fix web-ui console for bookworm

kergos:
Firstly the mtda-www package is added to the image. This is done because installing it manually afterwards with `sudo apt install mtda-www` from the fury repository installs it for python 3.9. and not the 3.11. version for bookworm.
Also, since the choice of using the webview comes from the configuration files anyway, it should not interfere.

Secondly, referring to #300, the `mtda/templates/index.html` and `mtda/assets/vnc.js` was changed to work with the newer novnc-package which is used in debian bookworm(https://packages.debian.org/bookworm/all/novnc/filelist). 

Consider that this will break the web-view for all mtda-images which are still built with bullseye.

### DIFF
--- a/meta-isar/recipes-core/images/mtda-image.bb
+++ b/meta-isar/recipes-core/images/mtda-image.bb
@@ -46,6 +46,7 @@ IMAGE_PREINSTALL += "                    \
     isc-dhcp-client                      \
     mjpg-streamer                        \
     mtda                                 \
+    mtda-www                             \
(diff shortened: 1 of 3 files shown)


chombourger (approved):
Thanks. I took note of the breakage for images based on bullseye. I can work on migrating those over to bookworm

Let's also plan on creating a new release soon to get packages hosted on fury to be updated

PR #312, opened 2023-08-04
remove references to bullseye, bookworm is now used in all configurat…

chombourger:
…ions


vj-kumar:
We might need to update the CI.
https://github.com/siemens/mtda/blob/d488beba8f33b560829048ea19298bcc4222f152/.github/workflows/main.yml#L56-L57

chombourger:
> We might need to update the CI.
> 
> https://github.com/siemens/mtda/blob/d488beba8f33b560829048ea19298bcc4222f152/.github/workflows/main.yml#L56-L57

good catch. I forgot to grep inside `.github`

vj-kumar (approved):

PR #316, opened 2023-08-04
check use of cache [do not review or merge]

chombourger:


chombourger:
Verified with https://github.com/siemens/mtda/actions/runs/5955475618
CI build went from 1h40m down to 10m for a PR that caused no bitbake signature changes at all (sstate fully used)

Issue #15, opened 2020-10-09
support bmap images


chombourger:
check if we can use the `BmapBdevCopy` from the `bmaptools` package to write images to the shared storage device
according to the documentation, the image object only needs to support `read()` and `seek()`

chombourger:
Merged (https://github.com/siemens/mtda/commit/01d7ef90b75e9b828060d48420f90571511c47f5)

PR #320, opened 2023-08-11
Added algorithm to auto-parse .bmap 2.0 files and write accordingly

kergos:
This update addresses issue #15.

Since the class `BmapBdevCopy` from the `bmaptools`  package expects a file stream to work properly and in this implementation we work with chunks of data, a custom algorithm for writing bmap 2.0 images is created.

In `client.py` a bmap file in the same folder as the image supplied with `mtda-cli storage write image.wic` is searched. If there is no .bmap  or the parsing of the .bmap file goes wrong, the image is written just like before. The algorithm was tested with 2 different images with holes. Each currently supported format (raw, bz, gz, xz and zst) was used for both of them and successfully flashed to a storage.

Credits go also to @fmoessbauer for supporting this endeavor.

### DIFF
--- a/mtda/main.py
+++ b/mtda/main.py
@@ -568,6 +568,17 @@ class MultiTenantDeviceAccess:
         self.mtda.debug(3, "main.storage_compression(): %s" % str(result))
         return result
 
+    def storage_bmap_dict(self, bmapDict, session=None):
+        self.mtda.debug(3, "main.storage_bmap_dict()")
+
+        self._session_check(session)
+        if self.storage_controller is None:
+            result = None
+        else:
+            self.storage_controller.setBmap(bmapDict)
+            result = True
+        self.mtda.debug(3, "main.storage_bmap_dict()(): %s" % str(result))
+
     def storage_close(self, session=None):
         self.mtda.debug(3, "main.storage_close()")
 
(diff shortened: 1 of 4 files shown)


chombourger:
Kudos to both of you for implementing this very much desired feature

kergos:
I forgot about the code formatting guidelines... sorry.
I will fix it up on Monday and issue another pull request.

chombourger:
No worries, we have these CI checks to catch such minor issues 😀

chombourger:
Please make sure to squash your 2 commits together (we'd like to avoid having a first commit and a second commit fixing the first)

mtda/main.py
```diff
@@ -1404,6 +1415,8 @@ def load_remote_config(self, parser):
                 # Load remote setting from the configuration
                 self.remote = parser.get(
                     'remote', 'host', fallback=self.remote)
+                # Allow override from the environment
+                self.remote = os.getenv('MTDA_REMOTE', self.remote)
```
chombourger, line 1419: do we really need this?
we recently added https://github.com/siemens/mtda/commit/e3b31cbe06b08241f15b72ccc53bc68e9f581bc7

kergos: Seems to be an artifact, I'll remove it. Sorry for the trouble

chombourger:
Looks good to me

chombourger (approved):

PR #334, opened 2023-09-13
feat: auto-detect serial interface of sdwire

fmoessbauer:
The serial of the sdwire interface depends on the physical device it is attached to. This makes it impossible to pre-configure multiple MTDA devices for a mass deployment.

This patch auto-detects the serial of an sdwire in case it is not specified in the configuration. This simplifies the default configuration and also enables pre-configuration of MTDA devices.

Note: This has been tested on multiple nanopi-neo boards with sdwire interfaces.

cc @bovi

### DIFF
--- a/docs/config.rst
+++ b/docs/config.rst
@@ -386,8 +386,8 @@ a SD card between the DUT and host. The following settings are supported:
   ``/dev/sda``)
 
 * ``serial``: string [optional]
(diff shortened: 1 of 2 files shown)


chombourger (approved):

chombourger:
that's a good idea. this will definitely improve the "out-of-the-box" experience. thank you!

PR #340, opened 2023-09-26
check for changes to systemd dropins

chombourger:


chombourger:
@fmoessbauer proposing these changes to only restart systemd when we need to
thoughts?

fmoessbauer:
I had exactly the same idea, but was afraid of the implementation effort (especially as we need to ignore the file timestamps). But your solution looks clean and straight forward.

While we are at this: Currently, pure dropin-style based configuration is not supported, as mtda always wants to have a config file in `/etc/mtda/config`. This could be fixed in this overhaul as well. Also, our own dropins should be prefixed with a number (e.g. `10-www.conf`), so users can easily override this, by using files starting at e.g. `90-<...>.conf`.

chombourger:
thank you @fmoessbauer for the feedback. I have raised #341 and #342 to address them (I usually like to make baby steps). I will probably start with #342 

I am not sure I fully understand #341 just yet

Issue #342, opened 2023-09-26
add a prefix (priority) to systemd drops


chombourger:
suggestion from @fmoessbauer (see #340)
having dropins prefixed with a number would make it easy for people to override settings with their own dropins

fmoessbauer:
Fixed in !344, resp. 534b9690003443428424089282c36168dc396b83

PR #343, opened 2023-09-27
Various fixes around service startup order

fmoessbauer:
For details, see individual commit messages. Especially have a look at the changes in p4.

cc @bovi

closes: #342


chombourger (approved):

fmoessbauer:
Looks like there was some change in the sd notify module. I'll push a fix.

fmoessbauer:
The package on pypi is called `systemd-python`. By that it should work with both the debian package, as well as the pypi package. @chombourger please trigger again.

chombourger (approved):

fmoessbauer:
@chombourger would you mind having a look at the tox errors. It looks like we need some native dependencies on the host which cannot be downloaded with pip. If it is too complicated, we could also make this conditional (by using a try import).

chombourger:
> @chombourger would you mind having a look at the tox errors. It looks like we need some native dependencies on the host which cannot be downloaded with pip. If it is too complicated, we could also make this conditional (by using a try import).

sure thing. I can take a look this evening

chombourger:
> > @chombourger would you mind having a look at the tox errors. It looks like we need some native dependencies on the host which cannot be downloaded with pip. If it is too complicated, we could also make this conditional (by using a try import).
> 
> sure thing. I can take a look this evening

finally got to it. reproduced the issue locally using WSL2+ubuntu-22.04 with systemd enabled (`systemd=true` in `/etc/wsl.conf'
fixed locally and now checking with #344 

if CI tests are now passing, we could then squash my commit into relevant commit(s) from this PR

chombourger:
Merged #344 hence closing this one.

PR #347, opened 2023-10-03
fix(video): avoid timeout during check for v4l2 device

fmoessbauer:
At least on debian bookworm, the v4l2 devices are not managed by systemd (the udev rule does not add the systemd tag). By that, we cannot use systemd to wait for the device as this will always timeout. It is also not easily possible to check if a device is (or will be) managed by systemd. Hence, we simply check if the device is present. In this case, we do not add the systemd device dependency. In other cases, we add it as the device might be managed by systemd and show up later.

We also considered adding a udev rule to tag the devices, but this makes the mtda package less portable.

cc @bovi

### DIFF
--- a/mtda/video/mjpg_streamer.py
+++ b/mtda/video/mjpg_streamer.py
@@ -54,6 +54,9 @@ class MJPGStreamerVideoController(VideoController):
 
     def configure_systemd(self, dir):
         device = os.path.basename(self.dev)
+        # the device is already there and might not be managed by systemd
+        if os.path.exists(device):
+            return
         dropin = os.path.join(dir, 'auto-dep-video.conf')
         with open(dropin, 'w') as f:
             f.write('[Unit]\n')


chombourger:
I suppose the udev approach would involve a rule such as the following:
```
KERNEL=="video0", SYMLINK="video0", TAG+="systemd"
```
and have the drop-in use the following clauses:
```
[Unit]
BindsTo=dev-video0.device
After=dev-video0.device
```
if so, what are your concerns around portability?

fmoessbauer:
> I suppose the udev approach would involve a rule such as the following:
> 
> ```
> KERNEL=="video0", SYMLINK="video0", TAG+="systemd"
> ```

In principle yes, but the `/dev/video*` device is not a symlink but the device itself. Maybe the following could work:

```
KERNEL=="video*", TAG+="systemd"
```

But I'm not sure how duplicated tags are handled. If that works, it could be a good solution. However, I'm unaware about why these devices are not managed by systemd in the first place, maybe there is a reason to it. How is it done on ubuntu? Do they have the udev rule?

chombourger:
looking into this

fmoessbauer:
Closing in favor of #353

PR #348, opened 2023-10-03
fix(client): clamp heartbeat to timeout value

fmoessbauer:
When using RPC timeouts shorter than the heartbeat value, the timeout of the RPC call was not properly working, as the call only returned after the heartbeat duration. In normal operation, this was not an issue, as the heartbeat was significantly shorter than the timeout, and also timeouts only happened due to function invocation timeouts, but not due to connectivity losses.

When working with one-shot RPCs (i.e. create a client to only call one rpc), this led to huge delays and an accumulation of half-open TCP connections. By clamping (limiting) the heartbeat duration to the timeout, also short timeouts are properly supported now.

### DIFF
--- a/mtda/client.py
+++ b/mtda/client.py
@@ -36,8 +36,9 @@ class Client:
         agent.load_config(host, config_files=config_files)
         if agent.remote is not None:
             uri = "tcp://%s:%d" % (agent.remote, agent.ctrlport)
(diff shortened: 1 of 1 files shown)


chombourger (approved):

fmoessbauer:
@chombourger can we merge this? Or is there still something missing?

PR #351, opened 2023-10-08
refactor(mtda): use upstream packages on bookworm

fmoessbauer:
Some packages we currently build locally are already available in debian bookworm. By using these, we automatically get security updates and also reduce the build time of the mtda images. This reduces the build time of the debian bookworm images by more than 10 minutes.

cc @bovi


chombourger:
I had local changes that also remove the recipes (as I don't think we need them anymore)

fmoessbauer:
> I had local changes that also remove the recipes (as I don't think we need them anymore)

Well... In ubuntu Jammy at least `python3-zstandard` is still missing: https://packages.ubuntu.com/search?keywords=zstandard&searchon=names

PR #349, opened 2023-10-03
fix(video): use udev rules to wait for the video device

chombourger:
systemd does not manage video device on its own but may be told to via udev rules that add "systemd" to the device TAGS. In addition, we need to resolve the name of the device in /dev to a sysfs device and then systemd .device unit.

PR #353, opened 2023-10-09
fix(video): avoid timeout during check for v4l2 device | v2

fmoessbauer:
This is another try at fixing the systemd unit dependency to the screen grabber. It is based on #349, but only uses the udev part. It has been tested on an nanopi-neo with a USB3 screen grabber attached.

cc @bovi


chombourger (approved):
This looks much cleaner than my PR, thanks!

PR #362, opened 2023-11-06
client: support 'storage write' from an S3 bucket

chombourger:
The 'storage write' command may now download an image (and its associated .bmap file if it exists) from an S3 bucket. Credentials may be provided in the environment (AWS_ACCESS_KEY_ID, AWS_SECRET_ACCESS_KEY).

### DIFF
--- a/debian/control
+++ b/debian/control
@@ -15,6 +15,7 @@ Package: mtda
 Architecture: all
 Multi-Arch: foreign
 Depends: python3:any (>= 3.7~),
+         python3-boto3,
          python3-daemon,
          python3-gevent,
          python3-libgpiod,
--- a/docs/usage.rst
+++ b/docs/usage.rst
@@ -108,7 +108,14 @@ command::
 
     $ mtda-cli storage write console-image.wic.bz2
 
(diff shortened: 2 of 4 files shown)


gylstorffq:
Would it be better to use HTTP/HTTPS instead? 
If you use AWS you can always access s3 with cloudfront and don't need to handle AWS credentials.

chombourger:
> Would it not be better to use HTTP/HTTPS instead? If you use AWS you can always access s3 with cloudfront and don't need to handle AWS credentials.

well I don't know how our admins have configured our s3 bucket but they have apparently disabled http/https (for sure the instance we use requires authentication)

chombourger:
> > Would it not be better to use HTTP/HTTPS instead? If you use AWS you can always access s3 with cloudfront and don't need to handle AWS credentials.
> 
> well I don't know how our admins have configured our s3 bucket but they have apparently disabled http/https (for sure the instance we use requires authentication)

I tried to access one of our images using https using wget --http-user --ask-password but without any success (getting a `Forbidden` error)

At any rate, I believe this change gives us an infrastructure to support more protocols and one could indeed add https.
Thoughts?

Issue #366, opened 2023-11-14
storage_update doesn't work 


DingMeng-Siemens:
_storage_update_ doesn't work  because:
1. the storage is not open before call __storage_write_ in _storage_update_
2. the __storage_write_ is removed from the latest commitment, but _storage_update_ still use this function

By the way, we have a use case: user just wants to flash the installer to the shared storage once, and copy some license files or different target images to installer frequently (there is a default target image in the /media partition which is used for user to copy their own images  in installer). But I don't see any solution for that situation.

bovi:
@chombourger it seems here the change was made: https://github.com/siemens/mtda/commit/0ccb5c6ba146d317940cdee1d5fcd48f58b9d517#diff-6a8cf01935ea3624a4ca3c07b29a8700aac7da76a1ce4d2932e6df875087eddaL179 and currently it is still used here: https://github.com/siemens/mtda/blob/master/mtda/client.py#L199

DingMeng-Siemens:
Already fix this,  I will make pull request in the next few days. cc @bovi

PR #368, opened 2023-11-17
storage: fix storage_update API

DingMeng-Siemens:
The storage_update need to change since the _storage_write is removed.
to solve issue #366 
cc @chombourger  @bovi

### DIFF
--- a/mtda/client.py
+++ b/mtda/client.py
@@ -182,21 +182,41 @@ class Client:
 
     def storage_update(self, dest, src=None, callback=None):
         path = dest if src is None else src
-        imgname = os.path.basename(path)
         try:
             st = os.stat(path)
-            imgsize = st.st_size
-            image = open(path, "rb")
+            image_size = st.st_size
         except FileNotFoundError:
             return False
 
         status = self._impl.storage_update(dest, 0, self._session)
         if status is False:
-            image.close()
             return False
+        blksz = self._agent.blksz
+        impl = self._impl
+        session = self._session
 
-        self._impl.storage_compression(CONSTS.IMAGE.RAW.value, self._session)
-        return self._storage_write(image, imgname, imgsize, callback)
+        # Get file handler from specified path
+        file = ImageFile.new(path, impl, session, blksz, callback)
+
+        # Open the shared storage device so we own it
+        self.storage_open()
+
+        try:
+            # Prepare for download/copy
+            file.prepare(image_size, CONSTS.IMAGE.RAW.value)
+
+            # Copy image to shared storage
+            file.copy()
+
+            # Wait for background writes to complete
+            file.flush()
+
+        except Exception:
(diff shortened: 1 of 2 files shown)


mtda/client.py
```diff
@@ -182,21 +182,42 @@ def storage_status(self):
...
-            image.close()
-            return False
 
-        self._impl.storage_compression(CONSTS.IMAGE.RAW.value, self._session)
-        return self._storage_write(image, imgname, imgsize, callback)
+        blksz = self._agent.blksz
+        impl = self._impl
+        session = self._session
+
+        file = ImageLocal(path, impl, session, blksz, callback)
```
chombourger, line 197: why not use `ImageFile` here so we can support both local and remote files?

chombourger, line 197: I am adding `ImageFile.new()` with https://github.com/siemens/mtda/pull/369
if used here, it will let us `storage update` a file which is either local or remote

mtda/client.py
```diff
@@ -182,21 +182,42 @@ def storage_status(self):
...
+
+        # Open the shared storage device so we own it
+        self.storage_open()
+
+        try:
+            # Prepare for download/copy
+            file.prepare(imgsize)
```
chombourger, line 204: with #369, `prepare()` can be told the compression format to use

DingMeng-Siemens: OK I will change.

mtda/client.py
DingMeng-Siemens: OK I will change.

chombourger: thank you. I have merged #369. you should be able to rebase your work

DingMeng-Siemens:
Refined the code.
I also removed _self._writer.start()_ from _storage_update()_ in main.py, because this will create an unmanaged writing thread and that will cause strange behavior since we already call _storage_open()_  in _storage_update()_ before writing in client.py. 
cc @chombourger  Please check.

mtda/client.py
```diff
@@ -182,21 +182,43 @@ def storage_status(self):
...
+        blksz = self._agent.blksz
+        impl = self._impl
+        session = self._session
+
+        # Get file handler from specified path
+        file = ImageFile.new(path, impl, session, blksz, callback)
+
+        # Open the shared storage device so we own it
+        # It also prevents us from loading a new bmap file while
+        # another transfer may be on-going
```
chombourger, line 203: this portion of the comment is not relevant (bmap files are only used with `storage write`)

chombourger:
Changes look good. just one comment needs to be updated

DingMeng-Siemens:
changed

chombourger (approved):

bovi:
@DingMeng-Siemens great work!

@chombourger thanks for merging

PR #371, opened 2023-11-22
fix some tricky bugs by setting flags at the proper timing and place

### LINKED ISSUE #267: Storage write ends with exception if image is compressed with xz or gz
If i write a image to a usb drive with mtda-0.17-1 on a raspi4 with `mtda-cli storage write <image>` the process ends with the following exception:

```
  File "/usr/bin/mtda-cli", line 825, in <module>
    app.main()
  File "/usr/bin/mtda-cli", line 814, in main
    status = cmds[cmd](stuff)
  File "/usr/bin/mtda-cli", line 473, in storage_cmd
    return cmds[cmd](args)
  File "/usr/bin/mtda-cli", line 563, in storage_write
    status = self.agent.storage_write_image(
  File "/usr/lib/python3/dist-packages/mtda/client.py", line 269, in storage_write_image
    return self._storage_write(image, imgname, imgsize, callback)
  File "/usr/lib/python3/dist-packages/mtda/client.py", line 210, in _storage_write
    status = self.storage_close()
  File "/usr/lib/python3/dist-packages/mtda/client.py", line 148, in storage_close
    return self._impl.storage_close(self._session)
  File "/usr/lib/python3/dist-packages/zerorpc/core.py", line 278, in <lambda>
    return lambda *args, **kargs: self(method, *args, **kargs)
  File "/usr/lib/python3/dist-packages/zerorpc/core.py", line 270, in __call__
    return self._process_response(request_event, bufchan, timeout)
  File "/usr/lib/python3/dist-packages/zerorpc/core.py", line 227, in _process_response
    reply_event = bufchan.recv(timeout=timeout)
  File "/usr/lib/python3/dist-packages/zerorpc/channel.py", line 255, in recv
    event = self._input_queue.get(timeout=timeout)
  File "src/gevent/queue.py", line 334, in gevent._gevent_cqueue.Queue.get
  File "src/gevent/queue.py", line 349, in gevent._gevent_cqueue.Queue.get
  File "src/gevent/queue.py", line 326, in gevent._gevent_cqueue.Queue.__get_or_peek
    SimpleQueue = _PySimpleQueue
  File "src/gevent/_waiter.py", line 151, in gevent._gevent_c_waiter.Waiter.get
  File "src/gevent/_greenlet_primitives.py", line 61, in gevent._gevent_c_greenlet_primitives.SwitchOutGreenletWithLoop.switch
  File "src/gevent/_greenlet_primitives.py", line 61, in gevent._gevent_c_greenlet_primitives.SwitchOutGreenletWithLoop.switch
  File "src/gevent/_greenlet_primitives.py", line 65, in gevent._gevent_c_greenlet_primitives.SwitchOutGreenletWithLoop.switch
  File "src/gevent/_gevent_c_greenlet_primitives.pxd", line 35, in gevent._gevent_c_greenlet_primitives._greenlet_switch
zerorpc.exceptions.LostRemote: Lost remote after 40s heartbeat

```

DingMeng-Siemens:
```sh

1. mtda-cli -r  $HOST_NAME storage write  1.0.1-installer.wic.xz
2. mtda-cli -r $HOST_NAME storage mount 3
3. mtda-cli -r $HOST_NAME storage update $image_path_in_installer.bmap $bmap_path
4. mtda-cli -r $HOST_NAME storage update $image_path_in_installer.xz $xz_path
5. mtda-cli -r  $HOST_NAME storage target
6. mtda-cli -r  $HOST_NAME target on
7. ...
8. mtda-cli -r  $HOST_NAME target off
9. mtda-cli -r  $HOST_NAME storage host
10. mtda-cli -r  $HOST_NAME storage write  1.0.2-installer.wic.xz
11. mtda-cli -r $HOST_NAME storage mount 3
```

When I run a shell script like above, I found some trick bugs: 

- The 1st commit to fix #267 **  . 
    I encountered the same error sometimes and the error message is just like this issue. 

   So the error in #267  means that there are still  many data to get but _file.flush()_ returned and running _storage_close()_  according to the storage write/update logic. And _storage close_ will wait for the writing thread join **until timeout.** 

    The reason why  _file.flush()_ returned it's the self._writing is **False** at the timeslot when the previous loop in the writing thread ended  but the next loop not started yet.

- The 2nd commit to fix the bug happening at **line 11** in the shell script.

   so we _storage mount_  firstly, and then we call _storage target_ which will umount the storage, but the self._storage_mounted is not set, so mount doesn't work anymore. 

   I add a static_value to set this flag, because it's not a session related, it might be umount by other session, so it should be a static variable in the Class instead of a Class Instance.

- The 3rd commit to fix the bug happening at **line 3** in the shell script.

   so we _storage write_  firstly, and then we call _storage update_. They are in the same session, so when calling _storage update_, the _storage_bmap_dict_ still have a value written by  _storage write_  last time, we  need to reset it after writing

cc @bovi  @chombourger

### DIFF
--- a/mtda/main.py
+++ b/mtda/main.py
@@ -643,7 +643,7 @@ class MultiTenantDeviceAccess:
         self.mtda.debug(3, "main.storage_mount()")
 
         self._session_check(session)
-        if self.storage_controller.static_is_mounted is True:
+        if self.storage.is_storage_mounted is True:
             self.mtda.debug(4, "storage_mount(): already mounted")
             result = True
         elif self.storage is None:
--- a/mtda/main.py
+++ b/mtda/main.py
@@ -643,7 +643,7 @@ class MultiTenantDeviceAccess:
         self.mtda.debug(3, "main.storage_mount()")
 
         self._session_check(session)
-        if self._storage_mounted is True:
+        if self.storage_controller.static_is_mounted is True:
             self.mtda.debug(4, "storage_mount(): already mounted")
             result = True
         elif self.storage is None:
--- a/mtda/storage/helpers/image.py
+++ b/mtda/storage/helpers/image.py
@@ -32,7 +32,8 @@ class BmapWriteError(OSError):
 
 
 class Image(StorageController):
-    static_is_mounted = False
+    _is_storage_mounted = False
+
     def __init__(self, mtda):
         self.mtda = mtda
         self.handle = None
@@ -45,6 +46,10 @@ class Image(StorageController):
         self.rangeChkSum = None
         self.lock = threading.Lock()
         atexit.register(self._umount)
+
+    @property
+    def is_storage_mounted(self):
+        return Image._is_storage_mounted
 
     def _close(self):
         self.mtda.debug(3, "storage.helpers.image._close()")
@@ -123,6 +128,7 @@ class Image(StorageController):
                     self.isloop = False
 
         self.mtda.debug(3, "storage.helpers.image._umount(): %s" % str(result))
+        Image._is_storage_mounted = result is False
         return result
 
     def umount(self):
@@ -130,8 +136,6 @@ class Image(StorageController):
 
         with self.lock:
             result = self._umount()
-        if result is True:
-            static_is_mounted = False
         self.mtda.debug(3, "storage.helpers.image.umount(): %s" % str(result))
         return result
 
@@ -229,7 +233,7 @@ class Image(StorageController):
         with self.lock:
             result = self._mount_impl(part)
         if result is True:
-            static_is_mounted = True
+            Image._is_storage_mounted = True
         self.mtda.debug(3, "storage.helpers.image.mount(): %s" % str(result))
         return result
 
@@ -298,7 +302,8 @@ class Image(StorageController):
         self.crtBlockRange = 0
         self.writtenBytes = 0
         self.overlap = 0
-        self.rangeChkSum = self._get_hasher_by_name()
+        if bmapDict is not None:
+            self.rangeChkSum = self._get_hasher_by_name()
 
     def supports_hotplug(self):
         return False
--- a/mtda/storage/helpers/image.py
+++ b/mtda/storage/helpers/image.py
@@ -32,7 +32,7 @@ class BmapWriteError(OSError):
 
 
 class Image(StorageController):
-
+    static_is_mounted = False
     def __init__(self, mtda):
         self.mtda = mtda
         self.handle = None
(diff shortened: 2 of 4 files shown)


chombourger:
flake8 complains about the following:
```
mtda/storage/helpers/image.py:36:5: E301 expected 1 blank line, found 0
mtda/storage/helpers/image.py:134:13: F841 local variable 'static_is_mounted' is assigned to but never used
mtda/storage/helpers/image.py:232:13: F841 local variable 'static_is_mounted' is assigned to but never used
```

DingMeng-Siemens:
> flake8 complains about the following:
> 
> ```
> mtda/storage/helpers/image.py:36:5: E301 expected 1 blank line, found 0
> mtda/storage/helpers/image.py:134:13: F841 local variable 'static_is_mounted' is assigned to but never used
> mtda/storage/helpers/image.py:232:13: F841 local variable 'static_is_mounted' is assigned to but never used
> ```

I will check on my local machine with flake8 :)

mtda/storage/helpers/image.py
```diff
@@ -32,7 +32,7 @@ class BmapWriteError(OSError):
 
 
 class Image(StorageController):
-
+    static_is_mounted = False
```
chombourger, line 35: Not sure I am liking the "static_" as a prefix because that's how class variables are implemented in c++ but the name for this concept is really class variables. I would simply omit that prefix. It is however good that you placed its initialization right after the class block and before any methods  because that's the usual practice

chombourger: And if it is private (which I believe it is), it should start with a leading underscore

mtda/main.py
```diff
@@ -643,7 +643,7 @@ def storage_mount(self, part=None, session=None):
         self.mtda.debug(3, "main.storage_mount()")
 
         self._session_check(session)
-        if self._storage_mounted is True:
+        if self.storage_controller.static_is_mounted is True:
```
DingMeng-Siemens, line 646: It's not private, we need to tell if it's mounted here.

mtda/storage/helpers/image.py
DingMeng-Siemens: > And if it is private (which I believe it is), it should start with a leading underscore

I will change it to is_mounted. But it's not private, because we use it to tell if it's mounted in 
https://github.com/siemens/mtda/pull/371/files#r1401607549

mtda/main.py
chombourger: ok I would then suggest we introduce a property. while _is_mounted should probably be a class variable for now, this could change in the future and users don't need to know

here's what I am suggesting
```
_is_mounted = False

@property
def is_mounted(self):
    return _is_mounted
```

how does that sound?

DingMeng-Siemens: sounds good :) but I will name it __is_storage_mounted_ and _is_storage_mounted_ to make it more clear :)

DingMeng-Siemens: updated, please check

PR #379, opened 2023-12-19
web-ui: add icon to toggle storage between host and target

fmoessbauer:


fmoessbauer:
@chombourger : Feel free to change the icons if you think the current ones are misleading.

cc @bovi

chombourger (approved):

chombourger:
> @chombourger : Feel free to change the icons if you think the current ones are misleading. Another thing to clarify is how to block the storage toggle option in the web UI in case the target is on. Shall we disable the button in this case, or just accept that the user can click it and nothing happens?
> 
> cc @bovi

probably a good idea for the backend to emit some storage-locked / storage-unlocked events but this can be done in a later PR in my opinion

PR #384, opened 2024-01-07
feature(www): use winbox for console & video panes

chombourger:


chombourger:
Screenshot from 2024-01-07 12-27-46

chombourger:
thoughts @fmoessbauer ?

fmoessbauer:
That's much better. In the past we often had issues when using both the serial as well as the screen, as both windows overlapped (and the z-index could not be changed).

chombourger:
Thanks.

PR #403, opened 2024-02-26
feat(www): Allow to use physical keyboard as hid input

TobiasSchaffner:
Import a pretty printed version of keysight.js from https://github.com/fresheneesz/keysight/blob/master/dist/keysight.umd.js and use it to catch the keyboard input in the web interface.

Adjust the keyboard mappings accordingly allow virtual and physical keyboard input to be forwarded to the hid input at the same time.


chombourger:
very cool. thank you!

chombourger (approved):

PR #404, opened 2024-02-26
power/gpio: Allow to configure low level triggered relays

TobiasSchaffner:
Introduce an optional `enable` config key for the power gpio. This config defaults to `high` for high level triggered relays but can be set to `low` for low level triggered relays.

### DIFF
--- a/mtda/power/gpio.py
+++ b/mtda/power/gpio.py
@@ -19,16 +19,28 @@ from mtda.power.controller import PowerController
 
 class GpioPowerController(PowerController):
 
+    HIGH = 1
+    LOW = 0
+
     def __init__(self, mtda):
         self.dev = None
         self.mtda = mtda
         self.lines = []
         self.gpiopair = []
+        self.trigger = self.HIGH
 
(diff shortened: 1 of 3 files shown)


mtda/power/gpio.py
```diff
@@ -19,17 +19,29 @@
...
 
         result = None
 
+        if 'enable' in conf:
+                    if conf['enable'] == 'high':
+                        self.trigger = self.HIGH
+                    elif conf['enable'] == 'low':
+                        self.trigger = self.LOW
+                    else:
+                        raise ValueError("'enable' shall be either 'high' or 'low'!")
```
chombourger, line 43: is it just the GitHub interface leading me to believe that hard tabs where used here instead of soft tabs (4 spaces)?

chombourger:
another very welcomed change/improvement. thanks!

mtda/power/gpio.py
TobiasSchaffner: Yes you are right. Fixed the indentation. Thanks!

Issue #20, opened 2020-10-09
compress image on-the-fly in storage write


chombourger:
The `storage write` command could compress image chunks on-the-fly to speed-up remote writes

PR #420, opened 2024-04-30
When copying uncompressed images, compress on them on-the-fly

fmoessbauer:
When copying an uncompressed image, we lightly compress the image on the fly with zstd on transfer. By that, subsequent zeros (holes) are automatically collapsed and not transferred, which can give a huge speedup on uncompressed images. Another advantage is, that zstandard can be decompressed with a very low overhead (e.g. compared to gz or xz) and is better suited for low-memory devices.

closes: #20

### DIFF
--- a/mtda/client.py
+++ b/mtda/client.py
@@ -17,6 +17,7 @@ import subprocess
 import tempfile
 import time
 import zerorpc
+import zstandard as zstd
 
 from mtda.main import MultiTenantDeviceAccess
 import mtda.constants as CONSTS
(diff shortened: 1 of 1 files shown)


fmoessbauer:
@chombourger This change is client-only, i.e. no need to update the backend. As I was not able to test the S3 path, please explicitly test this as well.

jan-kiszka:
Tested here over a low-speed link against a low-speed mtda device, and it works nicely.

jan-kiszka:
...except that zstandard here (0.13) does not work with `closefs=True`.

chombourger:
I haven't looked at your changes yet but 1st: thank you!
2nd: could you add a Closes: #20 line to your commit?

fmoessbauer:
Together with @jan-kiszka , we checked if the `closefs=True` is really needed (that was the reason for the zstandard version bump). However, the FD has always been closed (checked versions 0.14+), so we don't need to explicitly set it. By that, no need to update the minimal supported zstandard version.

chombourger (approved):

PR #426, opened 2024-10-04
feat(www): preliminary support for uploading to the shared storage

chombourger:


mtda/templates/index.html
```diff
@@ -235,11 +266,75 @@
...
+                uploadSelectedFile();
+              }
+            }
+            break;
           default:
             storage_status.innerHTML = "unknown_document"
             storage_status.title = "storage status unknown"
         }
       });
+      const CHUNK_SIZE = 512 * 1024;
```
fmoessbauer, line 282: What is the unit? Bytes?

fmoessbauer (commented):
In general this looks quite good (and I love to see that feature). I'm just wondering what happens if the upload is interrupted or gets stuck. Does this end-up in a non-recoverable situation as the storage remains open?

chombourger:
> In general this looks quite good (and I love to see that feature). I'm just wondering what happens if the upload is interrupted or gets stuck. Does this end-up in a non-recoverable situation as the storage remains open?

Reworked a session handling code a bit and will now close the shared storage device if the owning session has vanished

mtda/templates/index.html
chombourger: yes bytes. I need to test this feature a little more and in particular with large images. May even need to throttle the upload a bit as the browser will probably read the image much faster than we can write it. I have entered new territory while trying to implement this feature

PR #428, opened 2024-12-29
www upload and Ubuntu 24.04

chombourger:


fmoessbauer:
@chombourger Many thanks for adding this feature!

PR #431, opened 2025-01-06
Misc changes

chombourger:


kergoth (approved):
LGTM

chombourger:
> LGTM

thank you @kergoth

meta-isar/recipes-python/zerorpc/files/0002-do-not-use-future.patch
```diff
@@ -0,0 +1,75 @@
+License and copyright for files modified by this patch:
+
+SPDX-License-Identifier: MIT
+Copyright (c) 2015 François-Xavier Bourlet (bombela+zerorpc@gmail.com)
+
+Index: zerorpc-0.6.3/zerorpc/context.py
```
fmoessbauer, line 6: Does this mean the package is currently broken on trixie? Do we already have an upstream bug report?

chombourger: yes. I will submit the PR but the project appears to be orphaned. I have asked my colleagues in Bangalore to switch to a more main stream RPC framework (gRPC?)

chombourger: Ref: https://github.com/0rpc/zerorpc-python not touched for 4 years
I had filed a PR for a different issue/fix and it did not get the attention of the maintainers

Issue #376, opened 2023-12-11
Bad performance of storage write


fmoessbauer:
When flashing an image via `mtda-cli -r <host> storage write` to remove target, the speed of the copy is significantly lower than when first using scp and then `bmaptool copy` or `dd`. This is mainly due to the RPC based streaming of the image:

- serialization / deserialization overhead of the ~1MB chunks
- sequential RPC calls (Once the chunk copy command returns, the next is started. This makes each invocation delayed by at least one roundtrip time)
- Python GIL
  - The progress reporting calls interrupt the copy operations
  - The download operations interrupt the disk write operations

With the current zeroRPC based architecture there is not much room for improvement. Probably the best strategy would be to use zeroRPC only on the control path, but stream the image directly using zeromq. Also the receiver and the writer need to be put into two separate processes to parallelize this.

fmoessbauer:
With the on-the-fly zstd compression added in #420 , at least for uncompressed images the performance is much better now.

fmoessbauer:
With the recent switch to zmq for storage transfer (b116642df578bc5e01ed49c62ee6fc2e517c2e11), the performance is much better. Let's close this issue.

PR #434, opened 2025-01-07
replace zerorpc with pyro4

chombourger:
pyro4 is almost a dropin replacement for zerorpc (at least from a usage perspective). Further testing (fixing) is needed before we move forward.


chombourger:
some basic commands work but `storage write` fails:
```
an 07 17:15:40 DISW4XS7Q34 mtda-service[87575]: # debug1: storage.writer.worker(): a bytes-like object is required, not 'dict'
Jan 07 17:15:40 DISW4XS7Q34 mtda-service[87575]: # debug1: storage.writer.worker(): write or decompression error!
```
will have to investigate that failure

chombourger:
> some basic commands work but `storage write` fails:
> 
> ```
> Jan 07 17:15:40 DISW4XS7Q34 mtda-service[87575]: # debug1: storage.writer.worker(): a bytes-like object is required, not 'dict'
> Jan 07 17:15:40 DISW4XS7Q34 mtda-service[87575]: # debug1: storage.writer.worker(): write or decompression error!
> ```
> 
> will have to investigate that failure

using `marshal` as the serializer fixes that issue

chombourger:
ran several tests using mtda-kvm. will start testing using PIs (Raspberry PI4 and NanoPI NEO)

chombourger:
found on the nanopi that Pyro4 was listening on the loopback interface. added `Pyro4.config.HOST = "0.0.0.0"` to force listening for connections from any interfaces

fmoessbauer:
> seeing `storage write` being at least 2x slower with pyro4 compared to zerorpc. we may need to check configuration options for the pyro4 transport layer. alternatively: initiate the transfer using rpc and have the server open a socket that the client pay use to send the payload

The storage write is a bottleneck in general, as we pass the data via Python (discussed in #376). If we refactor this anyways, please try to avoid having Python in the data path.

chombourger:
if there are no objections, I would like to merge this to let people working with HEAD test, fix and/or enhance this initial port.

PR #436, opened 2025-01-13
Misc fixes

chombourger:


fmoessbauer:
The latest version of the kas-container script on kas:next has a fix for the Apparmor / bitbake issue. The container itself does not need to be changed.

PR #437, opened 2025-01-15
fix(docs): update config file name in install docs

ssnaaz:
The installation documentation needs to be updated to use the correct
 configuration file names for the Docker and KVM virtual platforms.
 The configuration files are now named mtda.ini instead of docker.ini
 and qemu.ini, respectively.


chombourger (approved):
thank you for your contribution!

PR #440, opened 2025-01-26
fix(storage): set high water marks for both rx&tx and retries for rx

chombourger:


chombourger:
with 64KiB buffers:
```
[####################] 100% (15.16 GiB read, 23.68 GiB written, 9.55 MiB/s)
```

chombourger:
tested by Shiva

PR #444, opened 2025-02-10
fix(client): don't allow storage write image and storage network simu…

ptsarath:
…ltaneously

Avoid conflict of flashing the image using systemtools like bmaptool or dd via network(nbd) and storage write of the image using the "storage write" command.

### DIFF
--- a/mtda/client.py
+++ b/mtda/client.py
@@ -9,7 +9,7 @@
 # SPDX-License-Identifier: MIT
 # ---------------------------------------------------------------------------
 
-
+import fcntl
 import os
 import Pyro4
 import random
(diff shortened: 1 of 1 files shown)


ptsarath:
@chombourger  Kindly look into the MR. Thank you.

mtda/client.py
```diff
@@ -26,6 +26,7 @@
 
 class Client:
 
+    _lockfile = "/tmp/storage_lock"
```
chombourger, line 29: shouldn't locking be done on the server side (main.py)? how are you handling PC-A and PC-B (different PCs) attempting to use the shared storage concurrently?

ptsarath: With the same PC but different terminal.

chombourger: we need to address the broader use-case where multiple clients attempt to use the shared storage

ptsarath: ok. We will check that as well. Thank you. On it !!

chombourger:
see #453

PR #461, opened 2025-03-13
A couple of improvements for the RPI4b target

fmoessbauer:
For details, see the commit messages.

Unfortunately for kernels < 6.5 we still can't use the DTB overlays, as the device tree of the `bcm2711-rpi*` targets does not contain the symbol information.

### DIFF
--- a/meta-isar/conf/machine/rpi4b.conf
+++ b/meta-isar/conf/machine/rpi4b.conf
@@ -13,7 +13,7 @@ IMAGE_FSTYPES ?= "wic"
 WKS_FILE ?= "rpi4b.wks"
 
 IMAGER_BUILD_DEPS = "rpi-firmware linux-image-${KERNEL_NAME}"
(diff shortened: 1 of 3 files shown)


fmoessbauer:
If we switch to the raspian kernel, we would have the information but other issues. I prefer to update to a 6.12 kernel later on and then continue the work on the overlays. This anyways needs to change again, once we switch to u-boot SPL booting to boot the target under EFI - which is needed to switch to the A/B rootfs boot via CIP. Tricky stuff.

chombourger (approved):

fmoessbauer:
> they pave the way to EFI boot

I would rather say they are mostly independent. I once implemented the EFI booting for the nanopi-neo target and that was a beast. By EFI booting, we completely leave the raspi boot path with the config files that are loaded by the firmware. Instead, all DTB fixups happen in uboot SPL. The good thing is that u-boot carries the device trees with symbol information, so we don't need to care about what the kernel offers (but we need to keep them at least loosely in sync).

PR #465, opened 2025-03-24
feat(rpi4): add target to support EFI booting

fmoessbauer:
As a preparation to support OTA updates via the swupdate pattern implemented in isar-cip-core, we enable EFI booting of the Raspberry Pi 4 B. The implementation boots the EDK2 variant for the RPi as a second stage, loads and patches the DTB and then performs standard EFI booting (using systemd-boot).
    
Due to limitations in the used EDK2 implementation, currently only booting from USB is supported. Booting from the SD card works as well, but is horribly slow as the controller runs the card in the "high-speed" mode during boot.


chombourger (approved):

chombourger:
I assume that we will eventually make this UEFI variant our primary target and drop the older one (to avoid maintaining 2 configurations of the rpi4b machine?)

chombourger:
reuse wants to add license/copyright information for
* meta-isar/conf/distro/debian-bookworm-backports.list
* meta-isar/conf/distro/preferences.mtda-bookworm-backports.conf
* meta-isar/recipes-bsp/rpi-edk2-firmware/files/config.override.txt
* patches/isar/0001-wic-add-IMAGE_EFI_BOOT_FILES-to-WICVARS.patch

fmoessbauer:
> I assume that we will eventually make this UEFI variant our primary target and drop the older one (to avoid maintaining 2 configurations of the rpi4b machine?)

Yes, I hope so, but I first want to have some more users on the EFI chain to carve out bugs. The current RPi4b target is anyways VERY limited, as none of the overlays works due to the kernel missing the DTB symbol information. On 6.12 this should be solved, but in my quick tests that did not solve the problem. On the EFI target, the overlays seem to work properly.

fmoessbauer:
> reuse wants to add license/copyright information for
> 
>     * meta-isar/conf/distro/debian-bookworm-backports.list
> 
>     * meta-isar/conf/distro/preferences.mtda-bookworm-backports.conf
> 
>     * meta-isar/recipes-bsp/rpi-edk2-firmware/files/config.override.txt
> 
>     * patches/isar/0001-wic-add-IMAGE_EFI_BOOT_FILES-to-WICVARS.patch

If you tell me how to add this information to these files, I'll do it. Does tox support external licensing, as I cannot add the license infos to e.g. the config file itself.

chombourger:
> If you tell me how to add this information to these files, I'll do it. Does tox support external licensing, as I cannot add the license infos to e.g. the config file itself.

sure thing. they may be added to https://github.com/siemens/mtda/blob/master/.reuse/dep5

PR #467, opened 2025-03-27
chore(docs): Add Raspberry Pi 4 Model B support

shikl3x7:
Add documentation to configure and use Raspberry Pi 4 Model B as the Assist Board.


docs/build.rst
```diff
@@ -158,6 +158,149 @@ and install the updated packages::
...
+    # vi /etc/mtda/config
+
+Hit ``i`` to enter the input mode and type the following configuration::
+
+    [console]
+    variant=usbf
+
+    [power]
+    variant=gpio
+    gpio=gpiochip0@529
```
fmoessbauer, line 285: This unfortunately currently neither works with the EFI target (as GPIO is not supported in EDK2) nor with the GPU boot target. I'm considering to switch the EFI target to u-boot as EFI provider to be able to pass the DT to Linux. As this DT is already compiled with symbols (`__symbol__` node), we should be able to apply the overlays.

I'm planning to implement this next week.

docs/RaspberryPi_Connections_MTDA.png
fmoessbauer, line 1: A special USB cable without VCC is needed to connect the USB-C port (gadet mode) to the DUT, as otherwise the DUT is powered via the 5V from the PI. This cable also can easily be made by just cutting the VCC.

docs/build.rst
chombourger: thanks for the info. I assume we can still go ahead with the documentation and that we will somehow find a way to make the GPIO stuff work? My Pi4 setup actually uses a USB relay. That should work with either boot modes I presume.

fmoessbauer: > I assume we can still go ahead with the documentation and that we will somehow find a way to make the GPIO stuff work

Yes, agree

> My Pi4 setup actually uses a USB relay

As this does not depend on a DT overlay, this should work.

chombourger (approved):

PR #468, opened 2025-03-27
fix(rpi): install correct device tree

fmoessbauer:
The imaging in ISAR happens in cross-mode. By that, the IMAGER_INSTALL packages are also installed in a cross chroot. However, the used apt pinnings only pin for the :native architecture. By that, the installed device tree is actually from a 6.1 kernel, while we run a 6.12 kernel in the image. This issue could be solved by pinning `linux-image-*:any`, however all this is not needed as we can just copy the device tree from the image rootfs. By that, we also can be sure that the DT matches the kernel.

Fixes: 1d3c93a ("feat: switch to 6.12 kernel from bookworm backports")


fmoessbauer:
... and we finally have a device tree with symbol information.

PR #470, opened 2025-03-31
A couple of usability improvements

fmoessbauer:


docs/config.rst
```diff
@@ -221,6 +221,15 @@ supports the following settings:
 * ``timeout``: integer [optional]
     Timeout (in seconds) for each connect.
 
+``hid`` driver settings
+~~~~~~~~~~~~~~~~~~~~~~~
+
+The ``hid`` driver emulates keyboard presses via the ``usbf`` driver:
+
+* ``device``: string [required]
+    Name of the kernel hid device (e.g. ``/dev/hidg0``). The device will be
```
fmoessbauer, line 230: Why isn't that the default in the keyboard driver? We could also set it there and just allow changing it here. Then it would become an optional parameter.

docs/config.rst
```diff
@@ -453,7 +462,8 @@ the DUT as a Mass Storage device. The following settings are supported:
 * ``file``: string [optional]
   File for the shared storage: a loopback device will be created to make
   sure that writes to the shared storage do not cause this file to
-  expand and leave the host without free space.
+  expand and leave the host without free space. This file needs to
+  be manually created (e.g. ``truncate -s 2G storage.img``).
```
fmoessbauer, line 466: From a usability PoV this is also not good. Should we add parameters to create this file with a fixed size? Can this file be on-the-fly compressed to save space (and reduce wearout of the SD card)? How about overlays?

docs/config.rst
chombourger: yes it sounds safe to make it the default

fmoessbauer: Ok, I'll add a commit that change this behavior and update the documentation.

docs/config.rst
chombourger: probably a good idea to do so indeed. I believe the qemu driver does this indeed for a better out-of-the-box experience

fmoessbauer: I aligned the new implementation with QEMU and also documented this. Please also test if this works for your workloads. Regarding the compression I was indeed looking into `dm-vdo`, but it was merely an idea to add this. Currently I'm not planning to actually implement it.

PR #474, opened 2025-04-04
Integrate swupdate with WFX

fmoessbauer:
We add a swupdate configuration to update the device with one of the DAU workflows from WFX. The WFX server can be specified with the WFX_SERVER bitbake variable (e.g. via a kas local config) and defaults to http://localhost:8080. By that, local tests with a SSH reverse forward can be done.

Once we have the general A/B rootfs image documented, we should also add example snippets to run the WFX workflow. But that's really just the example from https://siemens.github.io/wfx/docs/workflows/dau/

### DIFF
--- a/meta-isar-extra/recipes-conf/swupdate-config-wfx/swupdate-config-wfx_0.1.bb
+++ b/meta-isar-extra/recipes-conf/swupdate-config-wfx/swupdate-config-wfx_0.1.bb
@@ -1,0 +1,24 @@
+# This Isar layer is part of MTDA
+# Copyright (C) 2025 Siemens AG
+# SPDX-License-Identifier: MIT
+
+DESCRIPTION = "WFX backend configuration for swupdate"
+MAINTAINER = "mtda-users <mtda-users@googlegroups.com>"
+
+WFX_TENANT ??= "default"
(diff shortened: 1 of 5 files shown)


meta-isar-extra/recipes-conf/swupdate-config-wfx/swupdate-config-wfx_0.1.bb
```diff
@@ -0,0 +1,25 @@
+# This Isar layer is part of MTDA
+# Copyright (C) 2025 Siemens AG
+# SPDX-License-Identifier: MIT
+
+DESCRIPTION = "WFX backend configuration for swupdate"
+MAINTAINER = "mtda-users <mtda-users@googlegroups.com>"
+DPKG_ARCH = "all"
```
chombourger, line 7: this is now the default for `dpkg-raw` and it could be dropped (if you recommend we keep it explicit, that's also fine)

fmoessbauer: Ok, I will drop it and fix the missing license header.

chombourger (approved):

Issue #476, opened 2025-04-11
failure in storage write is not reported to client


fmoessbauer:
I just observed that a storage write that failed due to a bmap checksum error was not reported back to the client. On the server side, I got the following message in the log:

```
mtda-service[467]: # debug1:  ... mtda.storage.helpers.image.BmapWriteError: checksum mismatch for blocks range 688673-688673: calculated 598a6ad79d09d95c603f663339aedd134973fc8f22b0c06e0f8afa6fc3d0e4ac, should be 7387a340e9a0113b826fefa5ef20ebbc1a40ad617f58e3ad349faa39fc913628
mtda-service[467]: # debug1:  ...
mtda-service[467]: # debug1: storage.writer.worker(): write or decompression error!
```

I'm also wondering if there is a bug in our bmap algorithm implementation or if there was a transfer / decompression error before. I double checked that the same image+bmap can be flashed with the bmaptool.

PR #478, opened 2025-04-11
fix(image): correctly handle range end with overlap

fmoessbauer:
There was a subtle blug in our implementation of the bmap algorithm: When ending a range but still carying an overlap (a not fully written block), the next overlap is computed was computed incorrectly. By that, the writing position shifted by [0-blocksize) bytes which lead to incorrectly written data blocks. This was detected by a checksum mismatch.

We fix this by further simplifying the algorithm: Instead of computing the overlap based on a delta, we simply compute it by taking "writtenBytes % blocksize". We did this anyway in case the current input was fully processed, now we also do it within the block writing logic.

Xref: #476 (failure is still not reported, but the bug in writing is fixed).

Please review carefully and also test on some files.

### DIFF
--- a/mtda/storage/helpers/image.py
+++ b/mtda/storage/helpers/image.py
@@ -397,12 +397,11 @@ class Image(StorageController):
                              (cur_range["first"] - writtenBlocks)
                              * blksize - self.overlap)
                 self.handle.seek(nbytes, io.SEEK_CUR)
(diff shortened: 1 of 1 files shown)


chombourger (approved):

chombourger:
thank you @fmoessbauer - I will give it a try with various images. your notes really helped understanding the problem and your solution.

PR #480, opened 2025-05-14
chore(docs): Add Keyboard Usage Documentaion

shikl3x7:
Add documentation to use the Assist board as a keyboard to DUT.

### DIFF
--- a/docs/usage.rst
+++ b/docs/usage.rst
@@ -155,6 +155,64 @@ may be sent to the QEMU monitor using the ``command`` command::
 
     $ mtda-cli command hostfwd_add tcp::8080-:8080
 
+Keyboard
+~~~~~~~~
+
+The assist board has the capability. to act as keyboard to the DUT,general usage is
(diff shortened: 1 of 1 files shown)


docs/usage.rst
```diff
@@ -155,6 +155,61 @@ may be sent to the QEMU monitor using the ``command`` command::
 
     $ mtda-cli command hostfwd_add tcp::8080-:8080
 
+Keyboard
+~~~~~~~~
+
+The assist board has the capablity to act as keyboard to the DUT, using
```
chombourger, line 161: thanks for documenting the special keys. it may be a good idea to note that the general usage is `mtda-cli keyboard write characters...`. Some special characters are supported and need to be enclosed between `<` and `>`

chombourger: For instance: `mtda-cli keyboard write "<down><enter>hello world<enter>"`

chombourger:
see https://github.com/siemens/mtda/actions/runs/15039205489/job/42266750416?pr=480 for spellcheck errors

PR #481, opened 2025-05-17
ECM support

chombourger:


fmoessbauer:
Hi, that’s a nice feature. Does it also configure the kernel to allow routing (or natting) of the network?

chombourger:
> Hi, that’s a nice feature. Does it also configure the kernel to allow routing (or natting) of the network?

not yet. I'd like to add the following options:
- DHCP server
- IPSec
- NAT

Going step by step :)

chombourger:
Note: had tried that feature on the NanoPi NEO before but its USB controller would not have enough descriptors to let us create that many gadgets. Seems to work fine on the Pi4b

PR #482, opened 2025-05-19
Feat/network/usbf dhcp

chombourger:


chombourger:
@fmoessbauer next PR is probably going to add a mechanism to forward-NAT traffic. e.g. if I ssh to the Pi on port 2222, the rule would forward traffic to the DUT port 22. Is this how you also envisioned the NAT forwarding thing?

fmoessbauer:
> @fmoessbauer next PR is probably going to add a mechanism to forward-NAT traffic. e.g. if I ssh to the Pi on port 2222, the rule would forward traffic to the DUT port 22. Is this how you also envisioned the NAT forwarding thing?

I was actually thinking about just providing internet access to the DUT. But the ssh access is also important.

chombourger:
> > @fmoessbauer next PR is probably going to add a mechanism to forward-NAT traffic. e.g. if I ssh to the Pi on port 2222, the rule would forward traffic to the DUT port 22. Is this how you also envisioned the NAT forwarding thing?
> 
> I was actually thinking about just providing internet access to the DUT. But the ssh access is also important.

oh yes that as well. letting the DUT use the Pi as default gateway

PR #484, opened 2025-05-20
feat(network): add NAT forwarding

chombourger:
Let the agent forward traffic received on designated ports to the Device Under Test (which may be connected on a separate/isolated network).


chombourger:
@fmoessbauer this implements forwarding from the Pi to the DUT. Tested with SSH

chombourger:
added a commit to let us forward traffic from the DUT to the Internet as suggested by @fmoessbauer

PR #490, opened 2025-07-11
feat(storage): add initial support for commit/rollback to usbf

chombourger:


chombourger:
tested `storage` commands:
- commit
- rollback
- write

to be tested:
- mount
- update

PR #492, opened 2025-07-13
perf(cli): use event channel for "storage write" progress updates

chombourger:
When writing compressed images, a few reads may results in many writes: get progress updates from the event channel rather than client-side reads (via the progress callback).

### DIFF
--- a/mtda-cli
+++ b/mtda-cli
@@ -11,6 +11,7 @@
 # ---------------------------------------------------------------------------
 
 # System imports
+import os
 import requests
 import time
 import sys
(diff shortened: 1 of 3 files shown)


fmoessbauer:
Many thanks for implementing this. This is cleaner from an architectural PoV and also reports more evenly spaced back to the user. I'm just wondering if this breaks the ABI between the client and the server (i.e. using an older `mtda-cli` remotely and this one on the device). Do we have any checks for this, or at least state which versions are compatible?

Apart from that, we should also consider reporting progress based on what is actually written, instead of what is written + seeked (only if we have a bmap). This is also how bmaptool reports the progress.

PR #495, opened 2025-07-14
fix(storage): report actual bytes written (vs position)

chombourger:
Writes are not guaranteed to be sequential with bmap, we should therefore report bytes that were written to the storage instead of the position within the storage.

Suggested by: Felix Moessbauer <felix.moessbauer@siemens.com>

### DIFF
--- a/mtda/main.py
+++ b/mtda/main.py
@@ -1773,9 +1773,9 @@ class MultiTenantDeviceAccess:
         self.mtda.debug(4, f"main.notify: {result}")
         return result
 
-    def notify_write(self):
+    def notify_write(self, size=0):
         if self._writer:
-            self._writer.notify_write()
+            self._writer.notify_write(size=size)
 
     def start(self):
         self.mtda.debug(3, "main.start()")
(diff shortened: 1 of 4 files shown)


chombourger:
@fmoessbauer please check, this should address the issue you raised earlier in #491

fmoessbauer:
Thanks, I'll give it a try once I have updated my mtda RPIs (likely tomorrow).

fmoessbauer:
Hmm... this breaks on sparse images:

```
mtda-cli -r mtda storage write build/tmp/deploy/images/rpi4b-ebg/mtda-image-immutable-mtda-bookworm-rpi4b-ebg.wic
Discovered bmap file 'build/tmp/deploy/images/rpi4b-ebg/mtda-image-immutable-mtda-bookworm-rpi4b-ebg.wic.bmap'
mtda-image-immutable-mtda-bookworm-rpi4b-ebg.wic: [####################] 100% (374 MiB read, 496 MiB written, 10 MiB MiB/s) 
'storage write' failed! (image write failed: wrote 520885248 out of 4967121920 bytes)
```

The reason likely is that in case of bmap we need to check if we wrote all mapped blocks instead of the whole image size. Previously, we accounted the seeks as writes, hence the numbers were matching. But now we only consider the writes, hence the numbers are different.

I further noticed, that the progress bar still idles at around 1-4% and then immediately jumps to > 90%.

chombourger:
Closing in favor of #538

PR #509, opened 2025-07-21
Feat/www/python

chombourger:


chombourger:
> Hi @chombourger This looks like an interesting feature, but what I miss is the reasoning: Which use-cases do you envision where this helps? How should a user profit from having it?

Hi @fmoessbauer, the vision is provide test teams (including ours) a way to quickly prototype automated test-cases using the same APIs than our already existing Pytest APIs (https://github.com/siemens/mtda/blob/master/mtda/pytest.py) right from their browser without having to install anything.

Issue #501, opened 2025-07-18
inconsistent error handling of storage errors


fmoessbauer:
While debugging #495, I noticed that storage exceptions raised by the server are not reliably detected by the client:

- Errors that happen while the client is in `file.copy()` are not detected, because no error handlers are setup and the zmq connection eventually blocks (deadlocks) as the buffers are full. This can be reproduced by injecting a checksum error in a bmap file in one of the first chunks.
- Errors like a partial write (`'storage write' failed! (image write failed: wrote 520885248 out of 4967121920 bytes)`) are detected by the client only, hence from server perspective the write was successful, while it actually was corrupted.
- After finished storage writes, the "transmitting" panel in the UI remains open (and a new one is created on the next write). This indicates, that no final (`INITIALIZED` / `CORRUPTED`) event is emitted.

I propose to rework the storage API in the following way:

- All checks are performed server side
- The client registers to the server events and raises on error

PR #536, opened 2025-08-08
fix(storage): break on storage errors during transmission

fmoessbauer:
Errors that happened while writing to the storage remained unnoticed to the client in case they happened during data transmission: The error was detected on the receiving side, which closed the ZMQ socket. This resulted in an infinite block on the sending side.

We fix it by using the nonblocking send API. In case the sending fails because the queue is full, we check the storage status for write errors. As a full queue is not per-se an error (esp. on slow networks), we implement a simple exponential backoff algorithm and also only check the status if we can't send data for too long (to avoid frequent RPC calls).

Xref: #501


chombourger:
Good idea. I only wonder if we should reset backoff to zero if no errors are seen after a little while (possibly if time elapsed since the last backoff wait is >= backoff_wait)?

chombourger:
ah never mind. _write_to_storage is writing a single chunk of data. sorry for thinking aloud :)

PR #538, opened 2025-08-12
refactor(storage): report actual bytes written and seeked (vs position)

fmoessbauer:
Writes are not guaranteed to be sequential with bmap, we should therefore report bytes that were written to the storage instead of the position within the storage.

[Felix: also account seeks into writes to fullfill written==size assertion on client side. Downgrade to refactoring]

Suggested by: Felix Moessbauer <felix.moessbauer@siemens.com>

### DIFF
--- a/mtda/main.py
+++ b/mtda/main.py
@@ -1812,9 +1812,9 @@ class MultiTenantDeviceAccess:
         self.mtda.debug(4, f"main.notify: {result}")
         return result
 
-    def notify_write(self):
+    def notify_write(self, size=0):
         if self._writer:
-            self._writer.notify_write()
+            self._writer.notify_write(size=size)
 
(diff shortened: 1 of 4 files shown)


fmoessbauer:
This replaces #495. To remain API compatible, we also account seeks into the writes. But at least this is a start to get rid of the global storage state.

fmoessbauer:
@chombourger This should not have any API breaking changes (at least not compared to the previous MTDA version). I would like to merge this prior to bigger refactorings, as this helps to reduce the review effort.

fmoessbauer:
@chombourger I have one last commit that fixes the non-linearity of the progress reporting in an ABI compatible way. Shall I add it here as well, or do you prefer a follow up MR?

PR #542, opened 2025-08-20
Storage write improvements

fmoessbauer:
- make it more linear
- validate last range as well

closes: #501


fmoessbauer:
This has been tested with various images, incorrect last ranges and images with a total size that is not dividable by the blocksize. The bmap writing code is not altered, so on the "happy path" there is no change.

PR #548, opened 2025-11-20
fix: storage_flush called before write complete

shikl3x7:
storage_flush on client side is called before the write is completed and thread.join() in writer takes time, hence timeout observed in client side.

Tested-by: Vidyasagar G C <vidyasagar.gc@siemens.com>
Signed-off-by: Shivaschandra KL <shivaschandra.k-l@siemens.com>

### DIFF
--- a/mtda/client.py
+++ b/mtda/client.py
@@ -333,8 +333,11 @@ class ImageFile:
     def flush(self):
         # Wait for background writes to complete
         agent = self._agent
-
         self._socket.send(b'')
+        writing = True
+        while writing:
+            _, writing, written = agent.storage_status()
+            time.sleep(0.5)
         success = agent.storage_flush(self._totalsent)
         self._socket.close()
         self._socket = None


chombourger:
storage_flush is already called in flush. I however see that we have a forever loop in flush(). it waits for the server (the Pi) to report that it is no longer writing. I would be curious to understand if your Pi is able to complete its write queue? we probably want the forever loop on the client side to abort if we see the number of written bytes stalling while status remains writing

shikl3x7:
> storage_flush is already called in flush. I however see that we have a forever loop in flush(). it waits for the server (the Pi) to report that it is no longer writing. I would be curious to understand if your Pi is able to complete its write queue? we probably want the forever loop on the client side to abort if we see the number of written bytes stalling while status remains writing

Yes, that was the case, client side timed out, Writer was still writing, Writing completed and we were able to boot the image also.

chombourger:
> Nov 20 08:44:11 mtda mtda-service[2860]: # debug3: storage.writer.stop()
> Nov 20 08:44:11 mtda mtda-service[2860]: # debug2: storage.writer.stop(): waiting on thread...
> Nov 20 08:44:11 mtda mtda-service[2860]: # debug2: storage.writer.stop(): all done
> ```

do we have logs for when the issue occurred?
I recommend having both client and server-side logs

chombourger:
> I failed to reproduce the issue on 0.43 version. Since the master is still on 0.42.2 this change could work.

Raised https://github.com/siemens/mtda/pull/553 to bring 0.43 to master

fmoessbauer:
> the loop was removed with:
> 
> ```
> commit 830a9e7ba53f3fc0f889013f0fffbbcf1cb1310a
> Author: Felix Moessbauer <felix.moessbauer@siemens.com>
> Date:   Wed Aug 13 09:41:20 2025 +0200
> 
>     refactor(storage): drop useless loop around storage_status on flush
> 
>     The agent.storage_flush is a blocking call that waits for completion of
>     the storage write operation. By that, there is no need to continously
>     check for the storage status after returning.
> 
>     Signed-off-by: Felix Moessbauer <felix.moessbauer@siemens.com>
> ```
> 
> you are now re-introducing it.
> 
> please summarize your findings to bring @fmoessbauer up-to-speed

We now have the loop **before** the `storage_flush()`, which makes sense for progress reporting.

shikl3x7:
https://github.com/siemens/mtda/blob/b58a008c8c91fcbe49f3bf9e44eecee92a6a17a8/mtda/main.py#L969
If no increment in written at server aka main.py for every Pyro4 call, call storage_close explicitly Treating at as no progress.
If no increment in written at client side -> call storage_close explicitly.

chombourger:
pycodestyle checks fail:
```
mtda/client.py:337:1: W293 blank line contains whitespace
```

PR #549, opened 2025-11-25
qemu storage commit/rollback

chombourger:


chombourger:
includes the version bump commit from #547

mtda/storage/qemu.py
```diff
@@ -22,7 +22,8 @@ class QemuController(Image):
 
     def __init__(self, mtda):
         super().__init__(mtda)
-        self.file = "usb-storage.img"
+        self.file = "/var/lib/mtda/usb-shared-storage.img"
```
fmoessbauer: Why use an absolute path here instead of the relative one we had before?

PR #558, opened 2025-12-26
Update isar, cip-core to latest next and kas-container to 5.1

baprusty:


kas-container
```diff
@@ -27,8 +27,9 @@
 
 set -e
 
-KAS_CONTAINER_SCRIPT_VERSION="4.8.1"
+KAS_CONTAINER_SCRIPT_VERSION="5.1"
 KAS_IMAGE_VERSION_DEFAULT="${KAS_CONTAINER_SCRIPT_VERSION}"
+KAS_CONTAINER_IMAGE_DISTRO_DEFAULT="debian-bookworm"
```
fmoessbauer, line 32: This is different from the official kas-container script. Please add a comment explaining the reasoning and also mention it in the commit message.

jan-kiszka: This is the well-known sbuild bug. Should be fixed only by the next kas release which will pull Debian 13.3(?) that is supposed to receive the fix.

baprusty: Updated in the MR: https://github.com/siemens/mtda/pull/559
Thanks.

PR #563, opened 2026-01-14
web-ui: add support to flash with bmap + fixes

fmoessbauer:
The mtda-cli already supports to write the image using a bmap file. By that, writing images with large holes is speedup a lot. Further, the image data ranges are checksumed. We now implement the same feature for the web UI as well.

### DIFF
--- a/mtda-www
+++ b/mtda-www
@@ -719,7 +719,15 @@ class StorageOpenHandler(RemoteCallHandler):
             }
         }}}
 
-    async def _handle_remote_request(self, mtda):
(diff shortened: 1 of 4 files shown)


fmoessbauer:
Note, that the UI is still not hardened against any sort of user-error (e.g. flashing while the storage is attached to target, multiple clicks on upload, etc...). We should really improve at this frontier, as currently the chances to break MTDA by an erroneous click in the UI is quite high. Also, errors are not reported back to the user at all.

chombourger (approved):

Issue #524, opened 2025-07-31
replace pyro with varlink or gRPC


chombourger:
While the switch from ZeroRPC to Pyro was relatively easy, the Debian bookworm to trixie migration revealed a protocol breakage as we were forced to switch from Pyro4 to Pyro5. With our usage of the RPC being limited to control functions, I would propose that we use varlink. RFC.

fmoessbauer:
We just stumbled upon this again in an mtda lab that uses Pyro5 from trixie. If we want to support larger deployments, we need a stable ABI between the client and the server. I'm further proposing to make the RPC interface explicit, so the `mtda-cli` package and the `mtda` package can be split (currently you need system python bindings just to run the client).

An alternative would be to just provide a REST interface (as we already have for the UI) and use that for the CLI as well. Technically I don't see a reason against this.

bniedermayr:
> An alternative would be to just provide a REST interface (as we already have for the UI) and use that for the CLI as well. Technically I don't see a reason against this.

I would vote for this. 

Using REST would reduce the number of distinct ports to use. Especially interesting and sometimes crucial for setups behind a firewall (#562). IMHO the communication could then run over a single port where the desired functions (webUI, video stream, console etc.) could then be provided by different REST endpoints that might establish (if required) websockets between the client and server.

chombourger:
Done with 0.46. Closing

PR #564, opened 2026-01-21
Fix: mtda-rpi4b-ebg target

bniedermayr:
This fixes the mtda-rpi4b-ebg which can be used to integrated full swupdate support.

Multiple things were fixed:

- Update isar and isar-cip core to to their current master. Fixes the reference to stale commits (isar-cip-core next).
  **Note:It's not recommended to reference isar-cip-core next since it's getting rebased from time to time**
- Switch to debian-trixie: The reason for this was to update u-boot (u-boot-rpi) which contains important fixes for booting on the rpi4b platform.

### DIFF
--- a/kas/debian/mtda-rpi4b-ebg.yml
+++ b/kas/debian/mtda-rpi4b-ebg.yml
@@ -16,5 +16,5 @@ header:
   version: 17
   includes:
     - kas/common/base.yml
-    - kas/debian/bookworm-base.yml
+    - kas/debian/trixie-base.yml
     - kas/opt/ab-rootfs.yml
--- a/kas/common/base.lock.yml
+++ b/kas/common/base.lock.yml
@@ -3,4 +3,4 @@ header:
 overrides:
     repos:
         isar:
-            commit: 680c07ee072483329885ba08b7b2b1f763326dc8
+            commit: c92b7c95b0b867aa477396cb331c28601ce9e5ec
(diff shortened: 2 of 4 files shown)


kas/debian/mtda-rpi4b-ebg.yml
```diff
@@ -16,5 +16,5 @@ header:
   version: 17
   includes:
     - kas/common/base.yml
-    - kas/debian/bookworm-base.yml
+    - kas/debian/trixie-base.yml
```
fmoessbauer, line 19: This makes sense, but we should keep the whole MTDA layer on a single default distribution, as the protocol between `mtda-cli` and the server is incompatible between bookworm and trixie (i.e. a bookworm client cannot access a trixie server and vice versa). #524

bniedermayr: So should I look for another fix in order to stay on bookworm? Or do you mean to upgrade the layer to trixie?

fmoessbauer: The solution itself is fine, but I'm not sure if we should apply this fix just to the rpi4 target. In the end, that means if you build for rpi4 it will be trixie based, otherwise it will be bookworm based. Let's ask @chombourger for guidance.

chombourger: given what mtda remains under heavy development, it probably makes sense to get everything on trixie and remove support for bookworm. it may be difficult to support both Debian N and N-1. Would you agree @fmoessbauer and @bniedermayr ?

fmoessbauer: Agree.

chombourger (approved):

Issue #562, opened 2026-01-08
make operation behind gateway more user friendly


fmoessbauer:
When operating mtda behind a ssh jumphost, multiple ports need to be forwarded. It would be good, if the number of ports could be reduced, e.g. by using a multiplexer.

Currently we have:

- www->port 9080: http / web UI
- video->port 8080: video capture (jpeg stream)
- remote->control 5556: mtda control
- remote->console 5557: mtda console
- remote->data 5558 (default: dynamic):  mtda dataport (storage transfer)

The www and video port can be unified by using nginx as reverse proxy and a rewrite rule:

```
if ($request_uri ~ '\?action=stream') {
  set $adr http://$phost:8080;
}
```

bniedermayr:
Using [sshuttle](https://github.com/sshuttle/sshuttle) on the device running the mtda cli might help here.

chombourger:
this would get better with #569 : console and data ports were removed; we use the gRPC channel for control, console and data

PR #565, opened 2026-01-22
feat(mtda-cli): find all mtda agents in give ip pattern

shikl3x7:
This commit is just proposal to start discussions on having an Api to search all agents and their status, which would potentially help in automation labs with multiple setups to get the status of agent and use it to deploy.
For now, I have used the storage_status, if the idea seems valid and fit, would like work and add required Apis in main to make this feature more dependable.

### DIFF
--- a/mtda-cli
+++ b/mtda-cli
@@ -13,11 +13,16 @@
 # System imports
 import os
 import requests
+import re
 import time
 import sys
 import socket
 from argparse import ArgumentParser, RawTextHelpFormatter
-
+# Pyro
+try:
+    from Pyro5.compatibility import Pyro4
+except ImportError:
+    import Pyro4
 # Local imports
 from mtda.main import MultiTenantDeviceAccess
 from mtda.client import Client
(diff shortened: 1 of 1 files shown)


fmoessbauer:
Hi, IMHO this should not be an mtda feature, but instead rely on lava, Netbox or a custom fronted implementation. MTDA can assist by sending mdns broadcasts to make other participants aware of its address / existence.

I used MTDA interactively in a larger lab by putting an Nginx as reverse proxy in front of the nodes. The Nginx can also be exposed without providing direct access to the testlab. Example configuration for that was added by me in 1e383d5614dec56872b7af1e9eca93829ecd80fd. With #563 the UI is also almost feature complete, despite it is still tricky to use non interactively as the websocket interfaces are not documented (are not public interfaces).

In general, running MTDA in a bigger lab is still tricky due to #562 and #524. Once these are solved, we could even think of a MTDA proxy as dedicated application that just relies data to the correct (and discovered) MTDA device. But this should be a standalone application which is not part of MTDA itself (the same applies to the client as its source code is tightly coupled with the server).

chombourger:
a while back, support for ZeroConf was added. see https://github.com/siemens/mtda/blob/0615c477a8dfcf37c8ee82313762d470e383473c/mtda-service#L88

the feature may need to be documented & tested.

shikl3x7:
> a while back, support for ZeroConf was added. see
> 
> https://github.com/siemens/mtda/blob/0615c477a8dfcf37c8ee82313762d470e383473c/mtda-service#L88
> 
> the feature may need to be documented & tested.

Shall I change the implementation to use the ZeroConf and propose or shall we use this as part of the of our automation framework and just document here the usage.

PR #567, opened 2026-02-13
feat: capture screenshot from mtda

shikl3x7:
Add feature to capture screenshots from HDMI streamer on MTDA.

### DIFF
--- a/mtda-cli
+++ b/mtda-cli
@@ -17,6 +17,7 @@ import time
 import sys
 import socket
 from argparse import ArgumentParser, RawTextHelpFormatter
+from pathlib import Path
 
 # Local imports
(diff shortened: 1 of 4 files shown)


mtda-cli
```diff
@@ -88,6 +89,28 @@ class Application:
 
     def debug(self, level, msg):
         self.client().debug(level, msg)
+    
+    def capture_screen(self,args):
```
chombourger, line 93: this exposes implementation details. the cli should use Pyro to make a request and let the backend talk to the video controller

shikl3x7: Modified code, please check 👍

chombourger:
see https://github.com/siemens/mtda/pull/584

PR #570, opened 2026-03-20
chore: update layer dependencies

fmoessbauer:
This fixes the build issue of gnuefi on all targets using the efibootguard bootloader.


chombourger:
thank you!

chombourger (approved):

PR #575, opened 2026-04-09
make console auto-reconnect to our gRPC service

chombourger:


mtda-cli
```diff
@@ -39,11 +40,27 @@ class AppOutput(ScreenOutput):
...
+        domain = info[0] if info else ""
+
+        if domain == CONSTS.EVENTS.CONNECTION:
+            status = info[1] if len(info) > 1 else ""
+            if status == CONSTS.CONNECTION.LOST:
+                return
+            elif status == CONSTS.CONNECTION.RECONNECTING:
+                backoff = info[2] if len(info) > 2 else "?"
+                msg = f"\r*** connection lost, reconnecting in {backoff}s..."
+                msg = f"{msg:<60}"
```
fmoessbauer, line 53: Why limit this message as well?

mtda-cli
```diff
@@ -39,11 +40,27 @@ class AppOutput(ScreenOutput):
         self.app = app
 
     def on_event(self, event):
-        event = event.split()
+        info = event.split()
```
fmoessbauer, line 43: Can we use a named tuple for this? It's impossible to know which index refers to what.

mtda-cli
chombourger: since we are prepending the message with `\r`, I just wanted to make sure that we clear trailing characters on the current line

mtda-cli
chombourger: been thinking about your suggestion. you are definitely right that use of positional indexes is fragile and we should rework our code. as events are using different string formats, it is hard to come with name tuples that would work for all event types. we probably need an event string parser of some sort so we get an event dataclass with a known and well defined struct. I would suggest we implement that refactoring as a follow-up task. would you agree?

fmoessbauer: Ah ok. I was hoping that the format is static.

fmoessbauer (approved):

PR #577, opened 2026-04-09
feat(power/qemu): generate/use a virtual machine UUID

chombourger:
SMBIOS can expose a unique machine identifier for the Operating System to identify the machine it is running on. Make the qemu power controller generate/use a machine ID and pass it over to qemu via the -uuid option.

### DIFF
--- a/mtda/power/qemu.py
+++ b/mtda/power/qemu.py
@@ -20,6 +20,7 @@ import tempfile
 import threading
 import time
 import multiprocessing
+import uuid
 
 # Local imports
 from mtda.power.controller import PowerController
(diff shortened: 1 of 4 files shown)


mtda/power/qemu.py
```diff
@@ -79,6 +81,8 @@ def configure(self, conf):
             self.swtpm = os.path.realpath(conf['swtpm'])
         elif os.path.exists(self.swtpm) is False:
             self.swtpm = None
+        if 'uuid' in conf:
+            self.uuid = conf['uuid']
```
fmoessbauer: This uuid should be parsed and set to none if not valid.

chombourger: isn't it for QEMU to determine what is valid or not? I am happy to parse/validate (easy)

fmoessbauer: Please parse upfront and inform the user about the error (e.g. don't start MTDA). This is an easy error we can check upfront before even starting qemu.

chombourger: ready for review and for you to resolve this conversation if you are ok with the v2

fmoessbauer (approved):

PR #584, opened 2026-04-17
video snapshot

chombourger:


mtda-cli
```diff
@@ -674,10 +690,18 @@ class Application:
...
+        s.add_argument(
+            "-c", "--compress", action="store_true", default=False,
+            help="Compress whitespace and strip VT100 escape sequences",
+        )
         s = subsub.add_parser(
             "flush",
             help="Flush content of the console buffer",
         )
+        s.add_argument(
+            "-c", "--compress", action="store_true", default=False,
```
fmoessbauer: Why add the argument twice? I'm also a bit unsure about the name. This is not really compression, but rather `condense`.

mtda-cli
```diff
@@ -602,6 +618,26 @@ class Application:
...
+        return 1
+
+    def video_snapshot(self, args):
+        data, content_type = self.client().video_snapshot()
+        if data is None:
+            print("video snapshot not available", file=sys.stderr)
+            return 1
+        output = args.output
+        if output is None:
+            sys.stdout.buffer.write(data)
```
fmoessbauer, line 637: Please check upfront if the output is (not) connected to a TTY (`not sys.stdin.isatty()`).

mtda-cli
chombourger: changed to condense as suggested. both flush and dump accept this argument. refactored

mtda-cli
chombourger: great suggestion! thank you

mtda/constants.py
```diff
@@ -101,6 +101,10 @@ class USB:
     HID_MOUSE = "/dev/mtda-hid-mouse"
 
 
+class VIDEO:
+    SNAPSHOT_CHUNK_SIZE = 64*1024  # 64 KiB
```
fmoessbauer, line 105: Hmm... formats again. The `HIGH_WATER_MARK` below is in bytes, while this is in kB. Can we also use bytes here?

chombourger: yes `HIGH_WATER_MARK` is in bytes:
```
HIGH_WATER_MARK = 16*1024**2
```
but so is `SNAPSHOT_CHUNK_SIZE`:
```
SNAPSHOT_CHUNK_SIZE = 64*1024
```
am I misunderstanding your suggestion @fmoessbauer ?

fmoessbauer: sorry, I was distracted by the `**2`. The current code is already correct.

fmoessbauer (approved):

PR #586, opened 2026-05-28
mtda: use /var/lib/mtda as home directory for mtda user

bniedermayr:
The mtda debian package expects the mtda user's home directory at /var/lib/mtda, not /home/mtda. The mtda-service package already creates the mtda user with /var/lib/mtda as the home directory.

This also fixes a rebuild failure during image-postprocess stage when building the mtda-image from the sstate cache, where usermod errors out with:

  usermod: directory /home/mtda exists


fmoessbauer (approved):

fmoessbauer:
This breaks the swupdate image, as there we deploy an empty `/var` partition. We need a `systemd-tmpfiles.d` entry to populate it, link it or create empty.

PR #587, opened 2026-05-28
fix(power): fix libgpiod imports for V1 based systems

bniedermayr:
libgpio-v1 does not have a "real" python package but rather a single C binding library.

Actually the import statement "from gpiod.line import Direction, Value" should not work.
Starting with libgpiod-v2 imports like gpiod.* starts to work.

Tested on both bookworm and trixie.

### DIFF
--- a/mtda/power/gpio.py
+++ b/mtda/power/gpio.py
@@ -11,14 +11,16 @@
 
 # System imports
 import gpiod
-from gpiod.line import Direction, Value
-from operator import itemgetter
 
 # Local imports
 from mtda.power.controller import PowerController
 
 
-GPIOD_V2 = hasattr(gpiod, "chip")
+try:
+    from gpiod.line import Direction, Value
+    GPIOD_V2 = True
+except (ImportError, ModuleNotFoundError):
+    GPIOD_V2 = False
 
 
 class GpioPowerController(PowerController):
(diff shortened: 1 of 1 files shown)


fmoessbauer:
The commit makes sense, however I'm wondering how did you test this? Why did it work before (because on bookworm the old pattern was still valid)? ~Please also add a fixes tag. (it's already there)~

bniedermayr:
> The commit makes sense, however I'm wondering how did you test this? Why did it work before (because on bookworm the old pattern was still valid)? ~Please also add a fixes tag. (it's already there)~

I was just replacing the distro in mtda-rpi4b.yml ("kas/debian/trixie-base.yml" instead of "kas/debian/bookworm-base.yml").
So actually we could move to trixie in general for this target. 

The offending line is the import statement. Whereas "import gpiod" worked it should have stopped working as soon as the "from gpiod.line import Direction, Value" has been added on a bookworm based system.

In the error cases the logs show a 

> error: power could not be found/loaded!
 


1cce630 introduced that issue. Maybe this was only tested on trixie?

> Please also add a fixes tag. (it's already there)

Yes, it's there. But i just noticed that i had the signoff tag given twice :-)

bniedermayr:
This was my config

```
 [main]
debug = 1

[console]
variant = serial
port = /dev/ttyUSB0
rate = 115200

[power]
variant = gpio
gpio = gpiochip0@17
enable = high

[storage]
variant = usbf
device = /dev/sda
```

fmoessbauer (approved):

PR #593, opened 2026-06-09
fix(ro-image): create mtda homedir

fmoessbauer:
On ro-images we deploy an empty var partition. By that, the mtda homedir in /var/lib/mtda currently does not exist on ro images. To fix this, we just create the (empty) dir with a tmpfiles.d entry, similar to how it is done in isar-cip-core. If the directory is present, nothing is done.

Previously this was not needed, as the copy and creation of /home is automatically handled by isar-cip-core.

Fixes: 28f7d2f4 ("mtda: use /var/lib/mtda as home directory for ...")

CC @bniedermayr


fmoessbauer:
Locally tested.

chombourger (approved):

PR #598, opened 2026-06-22
fix(docs): power shellcmd return codes corrected

johann-pfefferl:
Fix the configuration description for the power » shellcmd action


chombourger:
can we change the target branch to next. we merge there first

PR #600, opened 2026-07-24
Uv migration

chombourger:


fmoessbauer:
To be frank, I'm not a fan of the `uv` migration: The `uv` tool itself is not packaged in Debian (there is an [ITP](https://packages.debian.org/source/sid/uv) and also partial packagings, but the command itself is still missing).

PR #601, opened 2026-07-24
feat(qemu): add mtda-cli firmware command to switch boot firmware

chombourger:
Add "mtda-cli firmware [efi|legacy]" to query or switch the boot firmware used by the qemu power controller, between OVMF (EFI) and SeaBIOS (legacy). Querying with no argument returns the current mode; other power controllers report "not supported".

Config gains optional [efi] and [legacy] sections under [power] to point at OVMF pflash files or a custom legacy BIOS image; flat pflash_ro/pflash_rw/bios keys directly under [power] keep working for backward compatibility. The active mode is persisted across mtda-agent restarts in /var/lib/mtda/qemu-firmware and defaults to efi unless only [legacy] is configured (overridable via [firmware] default=). Switching modes while qemu is already running restarts the qemu process itself so the new firmware takes effect immediately.

### DIFF
--- a/mtda/grpc/mtda.proto
+++ b/mtda/grpc/mtda.proto
@@ -231,6 +231,11 @@ message StorageToggleResponse {
 // Target / power
 // ---------------------------------------------------------------------------
 
+message TargetFirmwareRequest {
+    string mode     = 1;   // "efi" or "legacy"; ignored when get_only is true
+    bool   get_only = 2;
+}
+
 message TargetStatusResponse {
     string status = 1;
 }
@@ -354,6 +359,7 @@ service MtdaService {
     rpc StorageWrite       (stream StorageChunkRequest) returns (StorageWriteResponse);
 
     // --- Target ---
+    rpc TargetFirmware     (TargetFirmwareRequest)   returns (StringResponse);
     rpc TargetLock         (Empty)                   returns (BoolResponse);
     rpc TargetLocked       (Empty)                   returns (BoolResponse);
     rpc TargetOff          (Empty)                   returns (BoolResponse);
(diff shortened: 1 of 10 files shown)


mtda/grpc/mtda.proto
```diff
@@ -354,6 +359,7 @@ service MtdaService {
     rpc StorageWrite       (stream StorageChunkRequest) returns (StorageWriteResponse);
 
     // --- Target ---
+    rpc TargetFirmware     (TargetFirmwareRequest)   returns (StringResponse);
```
fmoessbauer, line 362: Why do we need to expose this to the client?

chombourger: for the `mtda-cli firmware` command so we can switch between efi and legacy (without having to edit e.g. `/etc/mtda/config`)

chombourger: (I am evidently not a big fan of legacy bios but we needed to test some images for older hardware and wanted an easy switch mechanism for our automated test bench. efi remains our preferred boot method.

fmoessbauer: Ok, makes sense. I did not consider this a use-case.

chombourger:
found an issue with consoles not being re-attached to the new qemu/kvm instance on restart. fixed now.

chombourger:
Proposing to rebase this MR after https://github.com/siemens/mtda/pull/603

chombourger:
@fmoessbauer @baprusty this is now rebased and ready for your review

fmoessbauer (approved):

baprusty (approved):

PR #603, opened 2026-08-04
mouse support for QEMU

chombourger:


mtda/power/qemu.py
```diff
@@ -26,6 +27,8 @@
 from mtda.power.controller import PowerController
 from mtda.utils import Size, System
 
+QMP_SOCKET = "/tmp/qemu-mtda.qmp"
```
fmoessbauer: Sockets should not be created under `/tmp`, but run or `XDG_RUNTIME_DIR`.

fmoessbauer:
Just out of curiosity: How is the performance of this implementation? Is this "good enough" to be exposed via the Web UI for interactive use (similar to PiKVM)?

chombourger:
> Just out of curiosity: How is the performance of this implementation? Is this "good enough" to be exposed via the Web UI for interactive use (similar to PiKVM)?

our web UI only sends events on a click - would otherwise be too slow but that would be worth looking into (as a separate task)

mtda/power/qemu.py
chombourger: looks like this was a pre-existing issue with this PR adding one more /tmp artefact. commit 1 is now removing all uses of /tmp from power/qemu and console/qemu.

chombourger:
@fmoessbauer is this now OK?

baprusty (approved):

fmoessbauer (approved):
Yes, looks good.

Issue #592, opened 2026-06-09
nbd-server.service fails to start with default config


fmoessbauer:
This service is auto-enabled when starting, but fails if the config is empty.
I *guess* this is a debian policy violation, as the service must not be enabled if it cannot start (which should be reported upstream).
For the meantime, I recommend to just disable the service and let mtda start/stop it on-demand.

As I'm not familiar with that component, can someone please check?

```
journalctl --boot -u nbd-server.service --no-hostname
Jun 09 07:51:40 systemd[1]: Starting nbd-server.service - Network Block Device server...
Jun 09 07:51:40 nbd-server[697]: Could not parse config file: The config file does not specify any exports
Jun 09 07:51:40 nbd-server[697]: No configured exports; quitting.
Jun 09 07:51:40 systemd[1]: nbd-server.service: Control process exited, code=exited, status=1/FAILURE
Jun 09 07:51:40 systemd[1]: nbd-server.service: Failed with result 'exit-code'.
Jun 09 07:51:40 systemd[1]: Failed to start nbd-server.service - Network Block Device server.
```

chombourger:
sure. I seem to remember this being an upstream issue (will need to double check). as a work-around, mtda could install a systemd drop-in to inject a `ConditionPathExists` check

fmoessbauer:
> as a work-around, mtda could install a systemd drop-in to inject a ConditionPathExists check

Yes, that would help.

chombourger:
Merged #611 - closing

PR #611, opened 2026-09-03
fix(nbd-server): only start server if config is provided by mtda

fmoessbauer:
The nbd-server is managed by mtda. However, the server fails to start if no config provides exports. By that, the overall system state becomes degraded if that feature of mtda is not in use.

We now provide a dropin that only enables the service if the file exists.

Closes: #592


chombourger:
sorry I missed to notice this PR. we would need a small edit to make the reuse check happy: meta-isar/recipes-conf/network/files/90-mtda-require-config.conf

fmoessbauer:
@chombourger I just added the licensing part and also added another minor fix regarding homedir creation (not worth it's own PR).

chombourger (approved):

